PR #30, opened 2020-05-30
more multi-play work

jiangzhengwenjz:
TODO: 
- [x] split another multi-play file in code_0800A8BC.s
- [ ] ~~decompile code_08030C94.s~~
- [x] decompile code_08019F28.s


jiangzhengwenjz:
Can't do better right now, unless we take effort to understand what the client side is doing and why we need multiple multi-boot programs. 
TODO: 
- [ ] split another multi-play file in code_0800A8BC.s
- [ ] decompile code_08030C94.s

jiangzhengwenjz:
@Cleverking2003 For lack of time, decomp of code_08030C94.s is postponed but it will be my next TODO.

PR #32, opened 2020-06-01
More random decomps

Solitai7e:


src/code_0800A868.c
```diff
@@ -31,3 +31,23 @@ s32 sub_0800A868(u8 playerId, struct Object *obj)
 	
     return var2;
 }
+
+void sub_0800A8BC(void)
+{
+    gUnk_03000004 = 0;
```
jiangzhengwenjz, line 37: Does `*(u32 *)(IWRAM + 4) = 0` match? We've seen the pattern elsewhere so the symbol probably doesn't exist

src/code_0800A91C.c
```diff
@@ -0,0 +1,23 @@
+#include "global.h"
+#include "functions.h"
+
+// TODO: define file boundaries
```
jiangzhengwenjz, line 4: Is there a reason for you to put the function in a separate file, rather than keeping it in `code_0800A868.c`?

Solitai7e: That's what I was going to do originally but then it turns out there's already an existing declaration in `functions.h`, and so after consulting @Cleverking2003 I was told to move it to it's own file. (and leave the declaration in `functions.h`)

Cleverking2003: It seems like I confused you a bit. I should have said you should have left definition in the file where it was before and not split it.

src/code_0800A868.c
```diff
@@ -1,6 +1,8 @@
 #include "code_0800A868.h"
 #include "functions.h"
 
+#include "gba/defines.h"
```
jiangzhengwenjz, line 4: I think it's already included in other header?

Solitai7e: Didn't realize it was `IWRAM_START` instead of `IWRAM` at first, my bad.

jiangzhengwenjz: :man_facepalming: my bad

jiangzhengwenjz:
@Solitai7e Note that I've squashed your commits, so you can do something like `git fetch upstream && git reset --hard upstream/master` to sync with upstream

PR #58, opened 2020-11-29
Add ObjectBase to Object2

Cleverking2003:


include/functions.h
```diff
@@ -53,21 +54,21 @@ u32 sub_0803DD58(u8);
...
 void sub_08063D98(struct Kirby *, __attribute__((unused)) s32);
-u16 sub_0806F780(struct Object2 *);
+u16 sub_0806F780(struct ObjectBase *);
 void sub_0806FE64(u16, struct Object2*);
 
 void sub_08070580(void);
 void sub_080708DC(struct ObjectBase*, struct Sprite*, u32, u16, u8, u16);
 void sub_08073D2C(struct Object2*);
-struct Object2 *sub_0807A7E8(struct Object2 *); // TODO: it returns (substruct of struct Object2 *) whose size is 0x7C
+struct Object3 *sub_0807A7E8(struct Object2 *);
```
UraYukimitsu, line 71: What's Object3?

UraYukimitsu (changes requested):

include/functions.h
jiangzhengwenjz: https://github.com/jiangzhengwenjz/katam/blob/objectbase/include/data.h#L122
unknown struct of size 0x7C

UraYukimitsu: The struct you linked is called Kirby3, and wouldn't that suggest that ObjectBase itself is 0x7C bytes long (maybe the field at 0x78 is of ObjectBase* type)?

UraYukimitsu (approved):
👌

include/functions.h
jiangzhengwenjz: I was absolutely blind -- the struct name is now fixed. 
ObjectBase is at most 0x78 bytes, see this https://github.com/jiangzhengwenjz/katam/blob/objectbase/src/giant_rocky.c#L149

PR #63, opened 2021-01-17
decompile some of dark_mind

jiangzhengwenjz:


src/dark_mind.c
```diff
@@ -0,0 +1,2218 @@
...
+            else
+                sub_08101968(r3);
+        }
+        else
+            sub_08101630(r3);
+        break;
+    }
+}
+
+#define Macro_08100BD0() \
```
Cleverking2003, line 403: I've seen this pattern in some objects, maybe we can use this macro in other files?

src/dark_mind.c
```diff
@@ -0,0 +1,2218 @@
...
+void sub_0810181C(struct DarkMind *r4)
+{
+    struct DarkMind *r5 = r4;
+
+    DarkMindSetFunc(r4, 0, sub_08101B28);
+    r4->unk0.base.xspeed = 0;
+    r4->unk0.base.yspeed = 0;
+    r4->unk0.base.flags |= 0x40;
+    r4->unk0.kirby3 = sub_0803D46C(&r4->unk0);
+    if (r4->unk0.base.x > r4->unk0.kirby3->base.base.x)
```
Cleverking2003, line 926: This pattern is also pretty frequent. Maybe we should create a macro for this (ObjChooseXDir)?

Cleverking2003 (approved):
Looks good

PR #68, opened 2021-04-19
more dark_mind

jiangzhengwenjz:


src/dark_mind.c
```diff
@@ -2553,3 +2558,381 @@ void sub_0810487C(struct DarkMind *sb)
...
+    r5->unk0.base.flags &= ~2;
+    r5->unk0.base.counter = 0;
+    r5->unk0.unk9E = 0;
+    r5->unk0.unk9F = 0;
+    r5->unk0.base.flags &= ~2;
+    if (r5->unk0.base.x > r5->unk0.kirby3->base.base.x)
+        r5->unk0.base.flags |= 1;
+    else
+        r5->unk0.base.flags &= ~1;
+    if (r5->unk0.base.x - r5->unk0.kirby3->base.base.x >= 0)
```
Cleverking2003, line 2578: Can we use abs() here?

Cleverking2003 (changes requested):

src/dark_mind.c
jiangzhengwenjz: > Can we use abs() here?

Can you show me how to do it?

Cleverking2003:
```c
if (abs(r5->unk0.base.x - r5->unk0.kirby3->base.base.x) > 0x3000) {
    // do something
}
```

Cleverking2003 (approved):

PR #74, opened 2021-06-12
more dark_mind

jiangzhengwenjz:


include/dark_mind.h
```diff
@@ -66,9 +66,9 @@ struct Object10 // struct DarkMindForm2?
     u32 unk104;
     u32 unk108;
     u32 unk10C;
-    u32 unk110;
-    u32 unk114;
-    u32 unk118;
+    struct Object2 *enemy3;
```
Cleverking2003, line 69: Why are they named in this order?

jiangzhengwenjz: I'm following the order of how they are generated. The counter is going down so it's reversed.

Cleverking2003 (approved):

PR #128, opened 2022-06-07
match master_sword.c

EpochFlame:


include/master_sword.h
```diff
@@ -0,0 +1,10 @@
+#ifndef GUARD_MASTERSWORD_H
+#define GUARD_MASTERSWORD_H
```
jiangzhengwenjz, line 2:
```suggestion
#ifndef GUARD_MASTER_SWORD_H
#define GUARD_MASTER_SWORD_H
```

src/master_sword.c
```diff
@@ -0,0 +1,28 @@
+#include "master_sword.h"
+#include "kirby.h"
+#include "functions.h"
+
+static void sub_08025434(struct Object2* arg0);
+static void sub_08025440(struct Object2* arg0);
+
+struct Object2* CreateMasterSwordNonCollectible(struct Object* arg0, u8 arg1) {
```
jiangzhengwenjz, line 8:
```suggestion
void* CreateMasterSwordNonCollectible(struct Object* arg0, u8 arg1) {
```
CreateXXX are placed in an array so we make all of them return void *.

jiangzhengwenjz (changes requested):

src/master_sword.c
EpochFlame: will do, thanks for the feedback

EpochFlame:
updated the create function accordingly

include/master_sword.h
```diff
@@ -0,0 +1,10 @@
+#ifndef GUARD_MASTER_SWORD_H
+#define GUARD_MASTER_SWORD_H
+
+#include "data.h"
+#include "global.h"
+#include "object.h"
+
+struct void* CreateMasterSwordNonCollectible(struct Object* arg0, u8 arg1);
```
jiangzhengwenjz, line 8:
```suggestion
void* CreateMasterSwordNonCollectible(struct Object* arg0, u8 arg1);
```

jiangzhengwenjz (approved):

PR #142, opened 2022-06-30
disassemble more multi_boot images

jiangzhengwenjz:


jiangzhengwenjz:
The other 4 `subgame_loader`s are similar to the disassembled `speed_eaters.gba` so they can share the same directory in the future. 
The same goes for the JP variants.

jiangzhengwenjz:
Ready for review. And I think I'm done with multi-boot for now.

jiangzhengwenjz:
Ready for review. Forgot to remove a temp file.

PR #191, opened 2023-05-08
more pause_fade

jiangzhengwenjz:
I didn't create a macro for `sub_08128F44` on purpose since I don't know how to do so properly. It's fake-matching, inconsistent and requires shared variables.


Cleverking2003 (approved):

jiangzhengwenjz:
```
1413968 total bytes of code
822064 bytes of code in src (58.1388%)
591904 bytes of code in asm (41.8612%)
```

PR #232, opened 2024-02-28
match sub_0813C064

jiangzhengwenjz:
Only a single function was decompiled since it took way too long.


Cleverking2003 (approved):

jiangzhengwenjz:
```
1413968 total bytes of code
938912 bytes of code in src (66.4026%)
475056 bytes of code in asm (33.5974%)
```

PR #299, opened 2025-05-17
Decompilation of code_08124BE0.s

Someone0123-pas:
**General Remarks:**
- As suspected, the pause screen files are very interconnected. I'll try to tackle it as follows: First decompiling all related .s files into respective, separate .c and .h files (i.e. `pause_area_map.s`, `pause_help.s`, the rest of `pause_world_map.s`, and perhaps `code_08123950.s`), then afterwards refactoring and resplitting into separate/merged files (and also rather not until then documenting functions, structs, defining incbins, etc.). I have some ideas already how the new file structure could look like, but I think it makes more sense to discuss this later when things are clearer.
- I have lightly documented some bits with comments, this is just to make the documentation step later easier (and because I found it interesting)
- Some code blocks feel oddly repetitive, so I presume those were originally preproc macros (for example, code_08124BE0.c:356 or :415) - I chose to leave it as is because I find redundancy for such little blocks alright and clearer than introducing a separate, abstracting macro; at least as of now.

**Misc. Questions:**
- I'm not that sure which #include-directives belong into the .h file and the .c file. For example, both code_08124BE0.c and .h reference stuff from pause_world_map.h, should the directive be in both files (and thus global.h, data.h and perhaps task.h are exceptions), or does it suffice to put it into .h alone?
- Do all (especially static) functions need to be prototyped at the beginning of a file (if not already in .h), or only where needed?


include/code_08124BE0.h
```diff
@@ -0,0 +1,28 @@
...
+#include "data.h"
+#include "global.h"
+#include "pause_world_map.h"
+#include "task.h"
+
+// TODO: Into pause_help.h
+struct Unk_08D6113C {
+    /* 0x0 */ void* unkSrc;
+    /* 0x4 */ void* unkDest;
+}; /* size = 0x8 */
```
jiangzhengwenjz, line 13:
```suggestion
struct Unk_08D6113C {
    /* 0x0 */ const u32* unkSrc;
    /* 0x4 */ u32* unkDest;
}; /* size = 0x8 */
```
The addresses must be aligned for CpuSet

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+    /* 0x3 */ u8 unk3;  // y
+}; /* size = 0x4 */
+
+extern const u16 gUnk_081E07FC[];
+extern const struct Unk_02021590 gUnk_08358D94[];
+extern const struct Unk_02021590 gUnk_08359BE8[];
+extern const struct Unk_08359C48 gUnk_08359C48[];
+extern const u8 gUnk_08359C88[];
+extern const u16 gUnk_08359E84[];
+extern const void* gUnk_08359EC4;
```
jiangzhengwenjz, line 57:
```suggestion
extern const u32 gUnk_08359EC4[];
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+    color = RGB_WHITE;
+    if (gMainFlags & 0x10000) {
+        LoadBgPaletteWithTransformation(&color, 0, 1);
+    }
+    else {
+        DmaCopy16(3, &color, gBgPalette, 2);
+        gMainFlags |= 1;
+    }
+
+    LZ77UnCompVram(&gUnk_08359EC4, (void*)VRAM);
```
jiangzhengwenjz, line 440:
```suggestion
    LZ77UnCompVram(gUnk_08359EC4, (void*)VRAM);
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+
+extern const u16 gUnk_081E07FC[];
+extern const struct Unk_02021590 gUnk_08358D94[];
+extern const struct Unk_02021590 gUnk_08359BE8[];
+extern const struct Unk_08359C48 gUnk_08359C48[];
+extern const u8 gUnk_08359C88[];
+extern const u16 gUnk_08359E84[];
+extern const void* gUnk_08359EC4;
+extern const struct Unk_08D6115C* const gUnk_08D6115C[];
+extern const struct Unk_0812F1C_78* const gUnk_08D61188[];
```
jiangzhengwenjz, line 59: It's interesting that you marked these as const here but `gUnk_08D6113C` as non-const. I wonder it affects matching.

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+    unk_0803C95C->unk8 |= 0x0180;
+    unk_0803C95C->unk4 = ~0;
+    unk_0803C95C->unk6 = ~0;
+
+    color = RGB_WHITE;
+    if (gMainFlags & MAIN_FLAG_BG_PALETTE_TRANSFORMATION_ENABLE) {
+        LoadBgPaletteWithTransformation(&color, 0, 1);
+    }
+    else {
+        DmaCopy16(3, &color, gBgPalette, 2);
```
jiangzhengwenjz, line 199:
```suggestion
        DmaCopy16(3, &color, gBgPalette, sizeof(color));
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+        gMainFlags |= MAIN_FLAG_BG_PALETTE_SYNC_ENABLE;
+    }
+    sub_0803D2A8(0, 0xff);
+}
+
+// TODO: Find type of struct/datatype with size 4
```
jiangzhengwenjz, line 205: Is the struct really used? :thinking:

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+        DmaCopy16(3, gUnk_08359E84, gBgPalette, 0x40);
+        gMainFlags |= 1;
+    }
+
+    color = RGB_WHITE;
+    if (gMainFlags & 0x10000) {
+        LoadBgPaletteWithTransformation(&color, 0, 1);
+    }
+    else {
+        DmaCopy16(3, &color, gBgPalette, 2);
```
jiangzhengwenjz, line 436:
```suggestion
        DmaCopy16(3, &color, gBgPalette, sizeof(color));
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+}; /* size = 0xC */
+
+struct Unk_08359C48 {
+    /* 0x0 */ u8 unk0;  // x
+    /* 0x1 */ u8 unk1;  // y
+    /* 0x2 */ u8 unk2;  // x
+    /* 0x3 */ u8 unk3;  // y
+}; /* size = 0x4 */
+
+extern const u16 gUnk_081E07FC[];
```
jiangzhengwenjz, line 51:
```suggestion
extern const u16 gUnk_081E07FC[0x80];
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+static inline void SpriteInit_08125690(u16 animId, u8 variant, u8 palId) {
+    struct Sprite sprite;
+    SpriteInitNoPointer2(&sprite, 0x06012000, 0x280, animId, variant, 0, 0xff, 0x10, palId, 0, 0, 0x81000);
+}
+
+void sub_08125690(void) {
+    u16 color;
+
+    if (gMainFlags & 0x10000) {
+        LoadBgPaletteWithTransformation(gUnk_081E07FC, 0x80, 0x80);
```
jiangzhengwenjz, line 416:
```suggestion
        LoadBgPaletteWithTransformation(gUnk_081E07FC, 0x80, ARRAY_COUNT(gUnk_081E07FC));
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+    }
+    else {
+        DmaCopy16(3, gUnk_081E07FC, gBgPalette + 0x80, 0x100);
```
jiangzhengwenjz, line 419:
```suggestion
        DmaCopy16(3, gUnk_081E07FC, gBgPalette + 0x80, sizeof(gUnk_081E07FC));
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+        gMainFlags |= 1;
```
jiangzhengwenjz, line 420:
```suggestion
        gMainFlags |= MAIN_FLAG_BG_PALETTE_SYNC_ENABLE;
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+    }
+
+    if (gMainFlags & 0x10000) {
+        LoadBgPaletteWithTransformation(gUnk_08359E84, 0, 0x20);
+    }
+    else {
+        DmaCopy16(3, gUnk_08359E84, gBgPalette, 0x40);
+        gMainFlags |= 1;
```
jiangzhengwenjz, line 428:
```suggestion
        gMainFlags |= MAIN_FLAG_BG_PALETTE_SYNC_ENABLE;
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+    }
+
+    color = RGB_WHITE;
+    if (gMainFlags & 0x10000) {
+        LoadBgPaletteWithTransformation(&color, 0, 1);
+    }
+    else {
+        DmaCopy16(3, &color, gBgPalette, 2);
+        gMainFlags |= 1;
```
jiangzhengwenjz, line 437:
```suggestion
        gMainFlags |= MAIN_FLAG_BG_PALETTE_SYNC_ENABLE;
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+
+    if (gMainFlags & 0x10000) {
+        LoadBgPaletteWithTransformation(gUnk_081E07FC, 0x80, 0x80);
+    }
+    else {
+        DmaCopy16(3, gUnk_081E07FC, gBgPalette + 0x80, 0x100);
+        gMainFlags |= 1;
+    }
+
+    if (gMainFlags & 0x10000) {
```
jiangzhengwenjz, line 423:
```suggestion
    if (gMainFlags & MAIN_FLAG_BG_PALETTE_TRANSFORMATION_ENABLE) {
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+        LoadBgPaletteWithTransformation(gUnk_08359E84, 0, 0x20);
+    }
+    else {
+        DmaCopy16(3, gUnk_08359E84, gBgPalette, 0x40);
+        gMainFlags |= 1;
+    }
+
+    color = RGB_WHITE;
+    if (gMainFlags & 0x10000) {
```
jiangzhengwenjz, line 432:
```suggestion
    if (gMainFlags & MAIN_FLAG_BG_PALETTE_TRANSFORMATION_ENABLE) {
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+    else {
+        DmaCopy16(3, &color, gBgPalette, 2);
+        gMainFlags |= 1;
+    }
+
+    LZ77UnCompVram(&gUnk_08359EC4, (void*)VRAM);
+    sub_081251F8();
+
+    SpriteInit_08125690(gUnk_08359BE8[0].animId, gUnk_08359BE8[0].variant, 0x8);
+    SpriteInit_08125690(gUnk_08358D94[gLanguage * 3].animId, gUnk_08358D94[gLanguage * 3].variant, 0x9);
```
jiangzhengwenjz, line 444: Does it match with 2D array?

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+        sub_08126898();
+    if (!*sub_08002888(2, 0xb, 0))
+        sub_081268E8();
+    if (!*sub_08002888(2, 0xc, 0))
+        sub_08126938();
+    if (!*sub_08002888(2, 0x3, 0))
+        sub_08126988();
+    if (!*sub_08002888(2, 0x4, 0))
+        sub_081269D8();
+    if (!*sub_08002888(2, 0xe, 0))
```
jiangzhengwenjz, line 476: use `SUB_08002888_ENUM_UNK_*`

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+
+static inline void SpriteInit_08125690(u16 animId, u8 variant, u8 palId) {
+    struct Sprite sprite;
+    SpriteInitNoPointer2(&sprite, 0x06012000, 0x280, animId, variant, 0, 0xff, 0x10, palId, 0, 0, 0x81000);
+}
+
+void sub_08125690(void) {
+    u16 color;
+
+    if (gMainFlags & 0x10000) {
```
jiangzhengwenjz, line 415:
```suggestion
    if (gMainFlags & MAIN_FLAG_BG_PALETTE_TRANSFORMATION_ENABLE) {
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+    /* 0x2 */ u8 unk2;  // x
+    /* 0x3 */ u8 unk3;  // y
+}; /* size = 0x4 */
+
+extern const u16 gUnk_081E07FC[];
+extern const struct Unk_02021590 gUnk_08358D94[];
+extern const struct Unk_02021590 gUnk_08359BE8[];
+extern const struct Unk_08359C48 gUnk_08359C48[];
+extern const u8 gUnk_08359C88[];
+extern const u16 gUnk_08359E84[];
```
jiangzhengwenjz, line 56:
```suggestion
extern const u16 gUnk_08359E84[0x20];
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+    if (gMainFlags & 0x10000) {
+        LoadBgPaletteWithTransformation(gUnk_081E07FC, 0x80, 0x80);
+    }
+    else {
+        DmaCopy16(3, gUnk_081E07FC, gBgPalette + 0x80, 0x100);
+        gMainFlags |= 1;
+    }
+
+    if (gMainFlags & 0x10000) {
+        LoadBgPaletteWithTransformation(gUnk_08359E84, 0, 0x20);
```
jiangzhengwenjz, line 424:
```suggestion
        LoadBgPaletteWithTransformation(gUnk_08359E84, 0, ARRAY_COUNT(gUnk_08359E84));
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+    }
+    else {
+        DmaCopy16(3, gUnk_08359E84, gBgPalette, 0x40);
```
jiangzhengwenjz, line 427:
```suggestion
        DmaCopy16(3, gUnk_08359E84, gBgPalette, sizeof(gUnk_08359E84));
```

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+// In code_08123950.s
+extern void sub_0812403C(struct Unk_0203ACC0*);
+
+// In pause_area_map.s
+extern u32 sub_08128694(u32);
+
+
+#define SpriteInitWorldMapObj(_sprite, _tilesVram, _unk14, _animId, _variant, _unk16, _unk1B, _unk1C, _palId, _x, _y, \
+                              _unk8)                                                                                  \
+    {                                                                                                                 \
```
jiangzhengwenjz, line 17: Does `({})` match?

src/code_08124BE0.c
```diff
@@ -0,0 +1,485 @@
...
+
+    sub_08155128(&unkMapSprite->unk0);
+    if (!(unkMapSprite->unk50 & 0x0002)) {
+        sub_08155128(&unkMapSprite->unk28);
+    }
+}
+
+void sub_081251F8(void) {
+    struct Sprite unkSprite;
+    CpuFill32(0, &unkSprite, 0x28);
```
jiangzhengwenjz, line 293:
```suggestion
    CpuFill32(0, &unkSprite, sizeof(struct Sprite));
```

jiangzhengwenjz (changes requested):

src/code_08124BE0.c
Someone0123-pas: Oh, I forgot to mark `gUnk_08D6113C` as const, it's matching as well.

Is my understanding correct that if I had defined it in some .c-file and still had forgotten to mark it as const, it wouldn't be linked into `.rodata`, right?

src/code_08124BE0.c
Someone0123-pas: That's a really good point! I've not managed to have `sub_08124F44` and the following `sub_08124F64` be called, and I can't find calls or their memaddress in Ghidra either.

I'll just rewrite the comment that these two funcs are probably unused.

src/code_08124BE0.c
Someone0123-pas: Yes, it matches, so I rewrote it.

This data label in particular is a bit of an odd case: As `NUM_LANGUAGES` is 6 and `sizeof(Unk_02021590)` is 4, one would expect it to have a range of 4 * 6 * 3 = 0x48 bytes.
It does however actually span a range of 0x5C data, this could perhaps be relevant later when defining it in the file, I think. And it also doesn't seem like this incbin is needed in any other file.

Someone0123-pas: image

Out of interest I looked into the byte structure, and indeed, after byte 0x48 the data looks to be very different from before and doesn't seem to be referenced anywhere.

I'll just add this as a comment for later documentation.

src/code_08124BE0.c
Someone0123-pas: Unfortunately not

Interestingly, I encountered that sometimes it really messed with codegen in this file, especially that issue where instead of just compiling to `mov rX, 0xff`, the compiler resorts to `mov rX, 1  ;  neg rX, rX` even though it only stores a byte - with macros enclosed as statement, it does the former more I think.
I encountered this for example when investigating how :356 behaves as macro, if I remember correctly.

src/code_08124BE0.c
jiangzhengwenjz: Right

src/code_08124BE0.c
jiangzhengwenjz: Great. I meant the struct was probably not used in sub_08124BE0 when I was writing the comment.

src/code_08124BE0.c
jiangzhengwenjz: It can be unref data and we can create a separate label for that when decompiling data.

jiangzhengwenjz (approved):

jiangzhengwenjz:
```
1413968 total bytes of code
1114500 bytes of code in src (78.8207%)
299468 bytes of code in asm (21.1793%)
```

PR #311, opened 2025-06-23
makefiles

ultra-azu:
- **Update tools**
- **Update makefile of main project**
- **Update makefiles of subgames_loader and its payload**
- **Do the rest of the Makefiles**
- **Force disable modern compilation**
- **Add nix flake**
- **Add charmap**
- **Fix make compare not comparing other subfolders, if not building**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,8 +156,9 @@ MAKEFLAGS += --no-print-directory
 # Delete files that weren't built properly
 .DELETE_ON_ERROR:
 
-RULES_NO_SCAN += clean clean-assets tidy tidymodern tidynonmodern
-.PHONY: all rom modern compare speed_eaters unk_8D94B9C unk_8E1FE28 unk_8E8490C
+# The subgames will build the tools if they aren't built anyway so it's fine
+RULES_NO_SCAN += clean clean-assets tidy tidymodern tidynonmodern speed_eaters unk_8D94B9C unk_8E1FE28 unk_8E8490C
+.PHONY: all rom modern compare
 .PHONY: $(RULES_NO_SCAN)
 
 infoshell = $(foreach line, $(shell $1 | sed "s/ /__SPACE__/g"), $(info $(subst __SPACE__, ,$(line))))
--- a/multi_boot/subgame_loaders/Makefile
+++ b/multi_boot/subgame_loaders/Makefile
@@ -155,8 +155,8 @@ MAKEFLAGS += --no-print-directory
 # Delete files that weren't built properly
 .DELETE_ON_ERROR:
 
-RULES_NO_SCAN += clean clean-assets tidy tidymodern tidynonmodern
-.PHONY: all rom modern compare payload
+RULES_NO_SCAN += clean clean-assets tidy tidymodern tidynonmodern payload
+.PHONY: all rom modern compare
 .PHONY: $(RULES_NO_SCAN)
 
 infoshell = $(foreach line, $(shell $1 | sed "s/ /__SPACE__/g"), $(info $(subst __SPACE__, ,$(line))))


ultra-azu:
It's worth mentioning I haven't deleted the `generated` stuff from pokeemerald, but it appears to be harmless. Also I didn't delete modern compilation, but instead modified a check to error out if someone specified. Seemed too much of a problem to delete everything about modern when we could have it in the future. If that's a problem I can delete it tho.

jiangzhengwenjz:
I must apologize that you probably misunderstood what I meant. I thought it would be a minimal patch with only the essential changes ported from pokeemerald. But of course a thorough revamp could be good.

I haven't reviewed the PR since it will certainly take a lot time, but I already have a few questions now:
1) I've made changes to preproc so it works without a charmap (since obviously katam doesn't need one). It looks like you just copied the charmap from pokeemerald which is useless here.
2) Would you mind explaining the Nix related files to me, like why they're necessary?
3) Does INSTALL.md need to be changed?

ultra-azu:
Yeah I think I went a bit overboard :sweat_smile: but at the end a bit of updating can't hurt.

1. Ah didn't knew that. I can revert the changes to preproc.
2. Not directly necessary (ig they could be kept separate on a Discord pin tbh). But it's an easy way for Nix/NixOS users to have all the tools needed right from the get go.
3. Ah, forgot about that. Only if the nix stuff gets accepted, in which case a line mentioning that they need flakes enabled and to type `nix develop` should be added.

ultra-azu:
I added your changes to preproc to the new preproc, that way the new preproc doesn't needs charmap.txt (I tried reverting to the old one but that didn't work properly, so that seemed like the best solution). I also kinda copy pasted from pokeemerald's INSTALL.md so I think that's good now too.

INSTALL.md
```diff
@@ -1,6 +1,6 @@
 First, you must put a Kirby & The Amazing Mirror (U) ROM (with SHA1: `274b102b6d940f46861a92b4e65f89a51815c12c`) in the root directory of the repository and name it `baserom.gba`. 
 
-## Prerequisites
+# Prerequisites
```
jiangzhengwenjz: Why would you want to change this?

charmap.txt
jiangzhengwenjz, line 1: We don't need a charmap

tools/preproc/asm_file.cpp
```diff
@@ -26,33 +26,13 @@
 #include "char_util.h"
 #include "utf8.h"
 #include "string_parser.h"
+#include "../../include/constants/characters.h"
```
jiangzhengwenjz: It doesn't build on my end.
> asm_file.cpp:29:10: fatal error: ../../include/constants/characters.h: No such file or directory

tools/preproc/preproc.cpp
```diff
@@ -123,57 +140,65 @@ char* GetFileExtension(char* filename)
...
     }
-
-    if (argc == 2
-        || (argv[2][0] == '-' && argv[2][1] == 'i' && argv[2][2] == '\0')) // ./preproc xxx.c or ./preproc xxx.c -i
+    if (optind + 1 == argc) {
         g_charmap = new (std::nothrow) Charmap("");
-    else
-        g_charmap = new (std::nothrow) Charmap(argv[2]);
-    if (!g_charmap)
-        FATAL_ERROR("Failed to allocate space for Charmap.\n");
```
jiangzhengwenjz: I think we can still keep the null check

Makefile
```diff
@@ -1,174 +1,297 @@
...
+  PATH_MODERNCC := PATH="$(PATH)" $(MODERNCC)
+  CC1 := $(shell $(PATH_MODERNCC) --print-prog-name=cc1) -quiet
+  override CFLAGS += -mthumb -mthumb-interwork -O$(O_LEVEL) -mabi=apcs-gnu -mtune=arm7tdmi -march=armv4t -fno-toplevel-reorder -Wno-pointer-to-int-cast
+  LIBPATH := -L "$(dir $(shell $(PATH_MODERNCC) -mthumb -print-file-name=libgcc.a))" -L "$(dir $(shell $(PATH_MODERNCC) -mthumb -print-file-name=libnosys.a))" -L "$(dir $(shell $(PATH_MODERNCC) -mthumb -print-file-name=libc.a))"
+  # LIB := $(LIBPATH) -lc -lnosys -lgcc -L../../libagbsyscall -lagbsyscall
+endif
+# Enable debug info if set
+ifeq ($(DINFO),1)
+  override CFLAGS += -g
+endif
```
jiangzhengwenjz, line 136: From what I see, -g is only missing when building modern. And IMO it should be on by default.

make_tools.mk
```diff
@@ -0,0 +1,24 @@
...
+# Use path relative to this makefile, in order fo Makefiles in subdirectory to work
+MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+TOOLS_DIR := $(dir $(MKFILE_PATH))/tools
+TOOL_NAMES := aif2pcm bin2c gbafix gbagfx mid2agb preproc scaninc
+
+TOOLDIRS := $(TOOL_NAMES:%=$(TOOLS_DIR)/%)
+
+# Tool making doesnt require a pokeemerald dependency scan.
+RULES_NO_SCAN += tools check-tools clean-tools $(TOOLDIRS)
+.PHONY: $(RULES_NO_SCAN)
```
jiangzhengwenjz, line 16: Aren't they already going to be set as phony targets in the main makefile?

make_tools.mk
```diff
@@ -0,0 +1,24 @@
+# This controls building executables in the `tools` folder.
+# Can be invoked through the `Makefile` or standalone.
+
+MAKEFLAGS += --no-print-directory
```
jiangzhengwenjz, line 4: Already in the main makefile?

make_tools.mk
```diff
@@ -0,0 +1,24 @@
...
+
+# Inclusive list. If you don't want a tool to be built, don't add it here.
+# Use path relative to this makefile, in order fo Makefiles in subdirectory to work
+MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+TOOLS_DIR := $(dir $(MKFILE_PATH))/tools
+TOOL_NAMES := aif2pcm bin2c gbafix gbagfx mid2agb preproc scaninc
+
+TOOLDIRS := $(TOOL_NAMES:%=$(TOOLS_DIR)/%)
+
+# Tool making doesnt require a pokeemerald dependency scan.
```
jiangzhengwenjz: Doesn't make sense in this repo

jiangzhengwenjz (changes requested):
I'll do further tests once you fix the build error.

ultra-azu:
Turns out I forgot to push some local changes I did a bit ago, my bad.

INSTALL.md
ultra-azu: I think I forgot the original and thought it looked better. I'll change it back though

tools/preproc/asm_file.cpp
ultra-azu: Yep. Forgot to commit that file. Fixed now

make_tools.mk
ultra-azu: I think that's done so that make_tools.mk does the same thing if it's invoked directly rather than by the main Makefile.

make_tools.mk
ultra-azu: If you mean it has the wrong name, done!

make_tools.mk
ultra-azu: Same reason as before, to allow using make_tools.mk on its own.

tools/preproc/preproc.cpp
ultra-azu: Done!

Makefile
ultra-azu: Done. What I did was set DINFO to 1 by default either on modern or nonmodern.

charmap.txt
ultra-azu: Done. That was the main change I didn't push oof

ultra-azu:
I think that's everything for now.

include/constants/characters.h
```diff
@@ -0,0 +1,339 @@
+#ifndef GUARD_CHARACTERS_H
```
jiangzhengwenjz: I wonder if this ought to be in include/ instead of tools/preproc/. After all our repo is not using the characters and the header is only for building the tool. Your thoughts? 
BTW, the header guard macro is inconsistent with those in other headers in constants/.

Makefile
```diff
@@ -187,47 +309,97 @@ multi_boot/unk_8E1FE28/unk_8E1FE28.gba: unk_8E1FE28
 multi_boot/unk_8E8490C/unk_8E8490C.gba: unk_8E8490C
 	@$(MAKE) -C multi_boot/unk_8E8490C COMPARE=$(COMPARE)
 
-$(C_BUILDDIR)/m4a.o: CC1 := tools/agbcc/bin/old_agbcc
+ifeq ($(MODERN),0)
+$(C_BUILDDIR)/m4a.o: CC1 := $(TOOLS_DIR)/agbcc/bin/old_agbcc
 $(C_BUILDDIR)/agb_sram.o: CFLAGS := -mthumb-interwork -Wimplicit -Wparentheses -Werror -O1 -g -fhex-asm -ffix-debug-line
-$(C_BUILDDIR)/agb_sram.o: CC1 := tools/agbcc/bin/old_agbcc
+$(C_BUILDDIR)/agb_sram.o: CC1 := $(TOOLS_DIR)/agbcc/bin/old_agbcc
 $(C_BUILDDIR)/task.o: CFLAGS := -mthumb-interwork -Wimplicit -Wparentheses -O2 -g -fhex-asm -f2003-patch -ffix-debug-line
```
jiangzhengwenjz: We can still use `O_LEVEL` here. The flag set is target-specific only because I had to turn off `-Werror`.

Makefile
```diff
@@ -187,47 +309,97 @@ multi_boot/unk_8E1FE28/unk_8E1FE28.gba: unk_8E1FE28
...
 	$(FIX) $@ -t"$(TITLE)" -c$(GAME_CODE) -m$(MAKER_CODE) -r$(REVISION) --silent
 
+# Builds the rom from the elf file
 $(ROM): $(ELF)
-	$(OBJCOPY) -O binary --gap-fill 0xFF --pad-to 0x9000000 $< $@
+	$(OBJCOPY) -O binary $< $@
+	$(FIX) $@ -p --silent
+
+# Symbol file (`make syms`)
+$(SYM): $(ELF)
```
jiangzhengwenjz, line 375: `*.sym` should be in .gitignore

Makefile
```diff
@@ -1,174 +1,296 @@
...
-clean-tools:
-	@$(foreach tooldir,$(TOOLDIRS),$(MAKE) clean -C $(tooldir);)
-
-clean: mostlyclean clean-tools
+clean-assets:
+	rm -f $(MID_SUBDIR)/*.s
+	rm -f sound/direct_sound_samples/*.bin
+	rm -f $(DATA_ASM_SUBDIR)/layouts/layouts.inc $(DATA_ASM_SUBDIR)/layouts/layouts_table.inc
+	rm -f $(DATA_ASM_SUBDIR)/maps/connections.inc $(DATA_ASM_SUBDIR)/maps/events.inc $(DATA_ASM_SUBDIR)/maps/groups.inc $(DATA_ASM_SUBDIR)/maps/headers.inc
+	find . \( -iname '*.1bpp' -o -iname '*.4bpp' -o -iname '*.8bpp' -o -iname '*.gbapal' -o -iname '*.lz' -o -iname '*.rl' -o -iname '*.latfont' -o -iname '*.hwjpnfont' -o -iname '*.fwjpnfont' \) -exec rm {} +
```
jiangzhengwenjz: It'd be better if we prepend `@` to all these lines. Currently there are way too many output lines when you run `make clean` considering there are subdirectories. And they're not helpful at all.

Same for other lines starting with `rm` and those in subdirectories.
(But brief and informative ones can still be there, like `@echo "rm -f <AUTO_GEN_TARGETS>"`)

Makefile
```diff
@@ -187,47 +309,97 @@ multi_boot/unk_8E1FE28/unk_8E1FE28.gba: unk_8E1FE28
...
+$(OBJ_DIR)/sym_bss.ld: sym_bss.txt
+	$(RAMSCRGEN) .bss $< ENGLISH > $@
+
+$(OBJ_DIR)/sym_common.ld: sym_common.txt $(C_OBJS) $(wildcard common_syms/*.txt)
+	$(RAMSCRGEN) COMMON $< ENGLISH -c $(C_BUILDDIR),common_syms > $@
 
-$(SONG_BUILDDIR)/%.o: $(SONG_SUBDIR)/%.s
-	$(AS) $(ASFLAGS) -I sound -o $@ $<
+$(OBJ_DIR)/sym_ewram.ld: sym_ewram.txt
+	$(RAMSCRGEN) ewram_data $< ENGLISH > $@
```
jiangzhengwenjz: We are not using ramscrgen.

multi_boot/unk_8D94B9C/Makefile
```diff
@@ -1,197 +1,353 @@
...
+# so you can't really call this rule directly
+generated: $(AUTO_GEN_TARGETS)
+	@: # Silence the "Nothing to be done for `generated'" message, which some people were confusing for an error.
 
-ifeq (,$(filter-out all rom compare,$(MAKECMDGOALS)))
--include tools
-endif
-
-baserom.gba: ../../baserom.gba
-	@dd if=$< of=$@ skip=14240668 count=125548 iflag=skip_bytes,count_bytes status=none
```
jiangzhengwenjz, line 139: It's impossible to build if you simply remove this. The `baserom.gba`s in subdirectories are generated based on the one in the root directory.

multi_boot/subgame_loaders/payload/Makefile
```diff
@@ -1,167 +1,354 @@
...
+# It doesn't look like $(shell) can be deferred so there might not be a better way (Icedude_907: there is soon).
+
+$(C_BUILDDIR)/%.o: $(C_SUBDIR)/%.c
+ifneq ($(KEEP_TEMPS),1)
+	@echo "$(CC1) <flags> -o $@ $<"
+	@$(CPP) $(CPPFLAGS) $< | $(PREPROC) -i $< ../../charmap.txt | $(CC1) $(CFLAGS) -o - - | cat - <(echo -e ".text\n\t.align\t2, 0") | $(AS) $(ASFLAGS) -o $@ -
 else
-$(C_BUILDDIR)/%.o: c_dep = $(shell $(SCANINC) -I include $(C_SUBDIR)/$*.c)
+	@$(CPP) $(CPPFLAGS) $< -o $*.i
+	@$(PREPROC) $*.i ../../charmap.txt | $(CC1) $(CFLAGS) -o $*.s
```
jiangzhengwenjz: charmap reference here

multi_boot/subgame_loaders/payload/Makefile
```diff
@@ -1,167 +1,354 @@
...
 
-ifeq ($(NODEP),1)
-$(C_BUILDDIR)/%.o: c_dep :=
+# As a side effect, they're evaluated immediately instead of when the rule is invoked.
+# It doesn't look like $(shell) can be deferred so there might not be a better way (Icedude_907: there is soon).
+
+$(C_BUILDDIR)/%.o: $(C_SUBDIR)/%.c
+ifneq ($(KEEP_TEMPS),1)
+	@echo "$(CC1) <flags> -o $@ $<"
+	@$(CPP) $(CPPFLAGS) $< | $(PREPROC) -i $< ../../charmap.txt | $(CC1) $(CFLAGS) -o - - | cat - <(echo -e ".text\n\t.align\t2, 0") | $(AS) $(ASFLAGS) -o $@ -
```
jiangzhengwenjz: charmap reference here

jiangzhengwenjz (changes requested):
I've found a few more issues. Some of my comments on a single makefile are actually applicable to all the makefiles.

include/constants/characters.h
ultra-azu: Done

Makefile
ultra-azu: Done

Makefile
ultra-azu: Done

Makefile
ultra-azu: Done

multi_boot/unk_8D94B9C/Makefile
ultra-azu: Done for all the sub-makefiles. I also added a rm -f for baserom.gba in those folders.

multi_boot/subgame_loaders/payload/Makefile
ultra-azu: Done

multi_boot/subgame_loaders/payload/Makefile
ultra-azu: Done

Makefile
ultra-azu: Done. Though the way I did it might be a bit opinionated.

jiangzhengwenjz:
Sure, please add those back.

multi_boot/subgame_loaders/payload/Makefile
```diff
@@ -1,167 +1,360 @@
...
+tidy: tidynonmodern tidymodern
 
-tidy:
-	@rm -f $(ROM) $(ELF) $(MAP)
-	@rm -r build/*
+tidynonmodern:
+	@rm -f $(ROM_NAME) $(ELF_NAME) $(MAP_NAME)
+	@rm -rf $(OBJ_DIR_NAME)
 
-include graphics_file_rules.mk
```
jiangzhengwenjz, line 114: graphics_file_rules.mk is removed from every makefile in subdirectories but the files are still there. I think we can still include them?

Makefile
```diff
@@ -38,52 +105,108 @@ DATA_ASM_BUILDDIR = $(OBJ_DIR)/$(DATA_ASM_SUBDIR)
...
+  override CFLAGS += -mthumb -mthumb-interwork -O$(O_LEVEL) -mabi=apcs-gnu -mtune=arm7tdmi -march=armv4t -fno-toplevel-reorder -Wno-pointer-to-int-cast
+  LIBPATH := -L "$(dir $(shell $(PATH_MODERNCC) -mthumb -print-file-name=libgcc.a))" -L "$(dir $(shell $(PATH_MODERNCC) -mthumb -print-file-name=libnosys.a))" -L "$(dir $(shell $(PATH_MODERNCC) -mthumb -print-file-name=libc.a))"
+endif
+# Enable debug info if set
+ifeq ($(DINFO),1)
+  override CFLAGS += -g
+endif
 
+# Variable filled out in other make files
+AUTO_GEN_TARGETS :=
```
jiangzhengwenjz: We can probably just remove everything about `AUTO_GEN_TARGETS` (like `generated`) since I can't think of anything in our repo that utilizes it.

multi_boot/unk_8D94B9C/Makefile
```diff
@@ -85,113 +214,147 @@ SONG_OBJS := $(patsubst $(SONG_SUBDIR)/%.s,$(SONG_BUILDDIR)/%.o,$(SONG_SRCS))
...
 
-tidy:
-	@rm -f $(ROM) $(ELF) $(MAP)
-	@rm -r build/*
+tidymodern:
+	@rm -f $(MODERN_ROM_NAME) $(MODERN_ELF_NAME) $(MODERN_MAP_NAME)
+	@rm -rf $(MODERN_OBJ_DIR_NAME)
 
-include graphics_file_rules.mk
-include songs.mk
```
jiangzhengwenjz, line 132: The reference is removed but the file is still there. It's different from the case of graphics_file_rules.mk since pokeemerald is no longer using it.

Makefile
```diff
@@ -1,33 +1,100 @@
-include $(DEVKITARM)/base_tools
+# GBA rom header
+TITLE       := AGB KIRBY AM
+GAME_CODE   := B8KE
+MAKER_CODE  := 01
+REVISION    := 0
+GAME_LANGUAGE := ENGLISH
+MODERN      ?= 0
```
jiangzhengwenjz: duplicate of line 16?

Makefile
```diff
@@ -187,47 +321,88 @@ multi_boot/unk_8E1FE28/unk_8E1FE28.gba: unk_8E1FE28
 multi_boot/unk_8E8490C/unk_8E8490C.gba: unk_8E8490C
 	@$(MAKE) -C multi_boot/unk_8E8490C COMPARE=$(COMPARE)
```
jiangzhengwenjz: I can't test modern at all but at least these targets are not using `MODERN` variable

Makefile
```diff
@@ -93,82 +216,93 @@ SONG_OBJS := $(patsubst $(SONG_SUBDIR)/%.s,$(SONG_BUILDDIR)/%.o,$(SONG_SRCS))
...
 sound/%.bin: sound/%.aif ; $(AIF) $< $@
 sound/songs/%.s: sound/songs/%.mid
 	cd $(@D) && ../../$(MID) $(<F)
 
+clean-generated:
+	@rm -f $(AUTO_GEN_TARGETS)
+	@echo "rm -f <AUTO_GEN_TARGETS>"
+
 # force targets
 speed_eaters: ;
```
jiangzhengwenjz: These should be phony targets (TBH I don't remember why I created these targets...)

Makefile
```diff
@@ -93,82 +216,93 @@ SONG_OBJS := $(patsubst $(SONG_SUBDIR)/%.s,$(SONG_BUILDDIR)/%.o,$(SONG_SRCS))
...
 
+# Other rules
 rom: $(ROM)
 ifeq ($(COMPARE),1)
-	@$(SHA1) katam.sha1
+	@$(MAKE) -C $(SUBGAME_LOADERS) compare
+	@$(MAKE) -C multi_boot/unk_8D94B9C compare
+	@$(MAKE) -C multi_boot/unk_8E1FE28 compare
+	@$(MAKE) -C multi_boot/unk_8E8490C compare
+	@$(SHA1) $(FILE_NAME).sha1
```
jiangzhengwenjz: This is definitely broken. If you do `make clean && make compare -j`, the checksum of payload.gba is going to be compared 3 times.

jiangzhengwenjz (changes requested):
Once again some of my comments are probably applicable to multiple makefiles.

multi_boot/subgame_loaders/payload/Makefile
ultra-azu: Done!

Makefile
ultra-azu: Done!

multi_boot/unk_8D94B9C/Makefile
ultra-azu: I think you meant deleting the songs.mk. If that's the case then done!

Makefile
ultra-azu: Yep. Done!

Makefile
ultra-azu: Done!

Makefile
ultra-azu: Done! I fixed it by not passing COMPARE to the other makefiles when building normally.

Makefile
ultra-azu: Done!

Makefile
```diff
@@ -93,141 +208,179 @@ SONG_OBJS := $(patsubst $(SONG_SUBDIR)/%.s,$(SONG_BUILDDIR)/%.o,$(SONG_SRCS))
...
-	@rm -f sound/direct_sound_samples/*.bin
-	@rm -f $(SONG_OBJS) $(MID_SUBDIR)/*.s
-	@find . \( -iname '*.1bpp' -o -iname '*.4bpp' -o -iname '*.8bpp' -o -iname '*.gbapal' -o -iname '*.lz' -o -iname '*.rl' -o -iname '*.latfont' -o -iname '*.hwjpnfont' -o -iname '*.fwjpnfont' \) -exec rm {} +
-	@rm -f $(AUTO_GEN_TARGETS)
-	@$(MAKE) -C $(SUBGAME_LOADERS) $@
+clean: tidy clean-tools clean-assets
+	@$(MAKE) -C multi_boot/subgame_loaders $@
 	@$(MAKE) -C multi_boot/unk_8D94B9C $@
 	@$(MAKE) -C multi_boot/unk_8E1FE28 $@
 	@$(MAKE) -C multi_boot/unk_8E8490C $@
```
jiangzhengwenjz: This is also broken. When `make clean`, the output is
```
Cleaning katam assets
rm -f sound/songs/midi/*.s
rm -f sound/direct_sound_samples/*.bin
find -iname '<graphics-related>' -exec rm {} +
Cleaning speed_eaters assets
rm -f sound/songs/midi/*.s
rm -f sound/direct_sound_samples/*.bin
find -iname '<graphics-related>' -exec rm {} +
rm -f baserom.gba
Cleaning unk_8D94B9C assets
rm -f sound/songs/midi/*.s
rm -f sound/direct_sound_samples/*.bin
find -iname '<graphics-related>' -exec rm {} +
rm -f baserom.gba
Cleaning unk_8E1FE28 assets
rm -f sound/songs/midi/*.s
rm -f sound/direct_sound_samples/*.bin
find -iname '<graphics-related>' -exec rm {} +
rm -f baserom.gba
Cleaning unk_8E8490C assets
rm -f sound/songs/midi/*.s
rm -f sound/direct_sound_samples/*.bin
find -iname '<graphics-related>' -exec rm {} +
rm -f baserom.gba
Cleaning speed_eaters assets
rm -f sound/songs/midi/*.s
rm -f sound/direct_sound_samples/*.bin
find -iname '<graphics-related>' -exec rm {} +
rm -f baserom.gba
Cleaning unk_8D94B9C assets
rm -f sound/songs/midi/*.s
rm -f sound/direct_sound_samples/*.bin
find -iname '<graphics-related>' -exec rm {} +
rm -f baserom.gba
Cleaning unk_8E1FE28 assets
rm -f sound/songs/midi/*.s
rm -f sound/direct_sound_samples/*.bin
find -iname '<graphics-related>' -exec rm {} +
rm -f baserom.gba
Cleaning unk_8E8490C assets
rm -f sound/songs/midi/*.s
rm -f sound/direct_sound_samples/*.bin
find -iname '<graphics-related>' -exec rm {} +
rm -f baserom.gba
```

multi_boot/subgame_loaders/Makefile
jiangzhengwenjz, line 1: This test fails:
1. Edit an instruction in `multi_boot/subgame_loaders/payload/asm/code.s` for a deliberate mismatch
2. `make compare -j` in ROOT_DIR
3. Change the instruction back to vanilla
4. `make compare -j` in ROOT_DIR (again)

Now at this point, the checksum will still fail.

This dependency issue also applies to other multi-boot subdirectories.

tools/preproc/asm_file.cpp
jiangzhengwenjz, line 1: Can you add `*.cpp text eol=lf` to .gitattributes?

Makefile
```diff
@@ -1,33 +1,99 @@
-include $(DEVKITARM)/base_tools
+# GBA rom header
+TITLE       := AGB KIRBY AM
+GAME_CODE   := B8KE
+MAKER_CODE  := 01
+REVISION    := 0
+GAME_LANGUAGE := ENGLISH
+KEEP_TEMPS  ?= 0
```
jiangzhengwenjz, line 7: How is this supposed to work? I tested with `make compare -j KEEP_TEMPS=1` and the `.i` and `.s` files are directly flooding into ROOT_DIR.

In addition, this is not used for the subdirectory targets.

Makefile
```diff
@@ -93,141 +208,179 @@ SONG_OBJS := $(patsubst $(SONG_SUBDIR)/%.s,$(SONG_BUILDDIR)/%.o,$(SONG_SRCS))
...
+$(ELF): $(LD_SCRIPT) $(OBJS)
+	@cd $(OBJ_DIR) && $(LD) $(LDFLAGS) -T ../../$< --print-memory-usage -o ../../$@ $(OBJS_REL) $(LIB) | cat
+	@echo "cd $(OBJ_DIR) && $(LD) $(LDFLAGS) -T ../../$< --print-memory-usage -o ../../$@ <objs> <libs> | cat"
 	$(FIX) $@ -t"$(TITLE)" -c$(GAME_CODE) -m$(MAKER_CODE) -r$(REVISION) --silent
 
+# Builds the rom from the elf file
 $(ROM): $(ELF)
-	$(OBJCOPY) -O binary --gap-fill 0xFF --pad-to 0x9000000 $< $@
+	$(OBJCOPY) -O binary $< $@
+	$(FIX) $@ -p --silent
```
jiangzhengwenjz, line 372: I understand this is a subtle change. But is it really beneficial to divide one command into two?

jiangzhengwenjz (changes requested):

multi_boot/subgame_loaders/Makefile
ultra-azu: Turns out moving the compare was a bad idea (makes sense, we want the build to quickly fail if the checksum doesn't match). Also makes sense in hindsight but the order of dependency for the subgames was wrong. So that's done

tools/preproc/asm_file.cpp
ultra-azu: Done!

Makefile
ultra-azu: Quite honestly, I'm not sure why it is the way it is. Flooding root is apparently the expected behavior (tested on pokeemerald) but I agree it's a bit ugly to do. I changed KEEP_TEMPS to have the original behavior of this project.

Makefile
ultra-azu: I think the idea is that the elf file has all the identifiers without ~~removing the debug symbols and the like that -p removes~~ gbafix doesn't do specifically that (that's what objcopy does), but I think the point still stands. The elf has all the identifiers already, so objcopy strips the debug symbols and gbafix pads it at the end.

ultra-azu: Ah wait, I think I understand now what you meant by looking at the original Makefile :woman_facepalming: Correct me if I'm wrong but manually padding like objcopy used to do will not necessarily lead to a power of 2 if someone modified katam.

Makefile
ultra-azu: I made it that way for it to be clear what subgame it's cleaning. But I can make it to output nothing on the subgames if you want.

jiangzhengwenjz: > s if you want.

No, I mean there are duplicates.

jiangzhengwenjz: > I made it that way for it to be clear what subgame it's cleaning. But I can make it to output nothing on the subgames if you want.

No, I mean there are duplicates.

Makefile
jiangzhengwenjz: Makes sense

Makefile
ultra-azu: Ah, I'm blind. Done by not calling clean directly in the other makefiles. Since clean-assets and tidy are called for them anw, it should be the same.

make_tools.mk
```diff
@@ -0,0 +1,24 @@
...
+# Use path relative to this makefile, in order fo Makefiles in subdirectory to work
+MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+TOOLS_DIR := $(dir $(MKFILE_PATH))/tools
+TOOL_NAMES := aif2pcm bin2c gbafix gbagfx mid2agb preproc scaninc
+
+TOOLDIRS := $(TOOL_NAMES:%=$(TOOLS_DIR)/%)
+
+# Tool making doesnt require a katam dependency scan.
+RULES_NO_SCAN += tools check-tools clean-tools $(TOOLDIRS)
+.PHONY: $(RULES_NO_SCAN)
```
jiangzhengwenjz, line 16: I don't really understand this part. For example, If I edit `tools/gbafix/elf.h`, the tool will not be rebuilt.

jiangzhengwenjz (changes requested):

make_tools.mk
ultra-azu: Might be the specific constant you modified. If I change `ELFMAG` from `"\177ELF"` to `"\177ELe"` a rebuild does get triggered. I'm guessing not all constants on that header are used

ultra-azu: Welp somehow I triggered it before but it doesn't now so nvm. It's weird though, if I change `gbafix.c` itself it does rebuild. Must be an issue with gbafix's Makefile

jiangzhengwenjz: Right. It seems other tools are handling header dependencies in their own makefiles so they don't require scaninc (since there's only one non-phony target to be built respectively). For some reason it's missing in gbafix's.

ultra-azu: Done! Ig that means pokeemerald has that bug too

jiangzhengwenjz: I guess so. You can submit a patch there. And probably another for fixing KEEP_TEMPS

jiangzhengwenjz (approved):

PR #313, opened 2025-07-01
Decompilation of pause_area_map.s

Someone0123-pas:
Some remarks:
- I've put declerations for `pause_area_map.c` functions into `pause_world_map.h` for the time being, so that they're all bundled for ease of later cleanlier file splitting/merging, etc.
- `sub_08128868` is most certainly a non-static inline, as I have seen it being inlined in `pause_world_map.s::sub_08127214`. As `sub_081286F0` is a non-static inline as well, `sub_08128788` and `sub_081287F4` in between must also be non-static inlines. For those however, I haven't seen them inlined, only passed as main-function (as destructors of the area map menue, so-to-speak).
- Immediately following are `sub_081288DC` and `sub_08128980`, which are then either also inlines or in a new file - As I have only gotten them called in subgames and not in the pause menue, I put them into `pause_fade.c`. Through that, I also assume that the whole of `pause_fade.c` is probably embedded in `subgames.c` and because it was so handy, the devs just utilised the screen fading functions from the subgames in the pause menues as well. This is something I'd plan to do later when restructuring the files.


include/pause_world_map.h
```diff
@@ -19,6 +19,22 @@
         }                                                                   \
     })
 
+#define SpriteInitMapObj(_sprite, _tilesVram, _unk14, _animId, _variant, _unk16, _unk1B, _unk1C, _palId, _x, _y, \
```
jiangzhengwenjz: Part of the `SpriteInit*` macros in sprite.h?

src/pause_area_map.c
```diff
@@ -0,0 +1,517 @@
...
+    gBgCntRegs[1] = BGCNT_PRIORITY(0) | BGCNT_CHARBASE(2) | BGCNT_16COLOR | BGCNT_SCREENBASE(23) | BGCNT_TXT256x256;
+    gBgCntRegs[2] = BGCNT_PRIORITY(1) | BGCNT_CHARBASE(0) | BGCNT_256COLOR | BGCNT_SCREENBASE(24) | BGCNT_TXT512x512;
+    gBgScrollRegs[0][1] = 8;
+    gBgScrollRegs[0][0] = 8;
+    gBgScrollRegs[1][1] = 0;
+    gBgScrollRegs[1][0] = 0;
+    gBgScrollRegs[2][1] = 0;
+    gBgScrollRegs[2][0] = 0;
+
+    task = TaskCreate(sub_08127FCC, sizeof(struct AreaMap), 0x1000, TASK_x0004, NULL);
```
jiangzhengwenjz:
```suggestion
    task = TaskCreate(sub_08127FCC, sizeof(struct AreaMap), 0x1000, TASK_USE_IWRAM | TASK_x0004, NULL);
```

src/pause_area_map.c
```diff
@@ -0,0 +1,517 @@
...
+        DmaCopy16(3, gUnk_08D6126C, gBgPalette + 0x50, sizeof(gUnk_08D6126C));
+        gMainFlags |= MAIN_FLAG_BG_PALETTE_SYNC_ENABLE;
+    }
+
+    CpuCopy32(gUnk_0835AECC, (void*)VRAM, sizeof(gUnk_0835AECC));
+    CpuCopy32(gUnk_08D616E4, (void*)0x06002000, sizeof(gUnk_08D616E4));
+    CpuCopy32(gUnk_08D612E4[0], (void*)0x06002400, sizeof(gUnk_08D612E4[0]));
+
+    areamap->unk5E = ~0;
+    areamap->unk5C = ~0;
```
jiangzhengwenjz:
```suggestion
    areamap->unk5E = -1;
    areamap->unk5C = -1;
```

src/pause_area_map.c
```diff
@@ -0,0 +1,517 @@
...
+            }
+        }
+    }
+
+    if (areamap->unk120[gUnk_0203AD3C].unk28 && (gUnk_0203ACC0[gUnk_0203AD3C].unkE & (0x0200 | 0x0100))) {
+        if (gUnk_0203ACC0[gUnk_0203AD3C].unkE & 0x0200) {
+            areamap->unk47 = 0x01;
+        }
+        else {
+            areamap->unk47 = 0xff;
```
jiangzhengwenjz:
```suggestion
            areamap->unk47 = -0x01;
```

src/pause_area_map.c
```diff
@@ -0,0 +1,517 @@
...
+                break;
+            }
+        }
+#else
+        ++areamap->unk5C;
+        while (gUnk_08D61B20[areamap->unk5C] == 0xff)
+            areamap->unk5C = 0;
+        CpuCopy32(gUnk_08D612E4[gUnk_08D61B20[areamap->unk5C]], (void*)0x06002400, sizeof(gUnk_08D612E4[0]));
+        ++areamap->unk5C;
+        areamap->unk5E = gUnk_08D61B20[areamap->unk5C];
```
jiangzhengwenjz: Do you think this is functionally equivalent? I want to have your opinion on it.

jiangzhengwenjz (changes requested):
I merged master into your branch and changed the parameter type of the lz77 uncomp functions so please pull before making any changes.

include/pause_world_map.h
soiryl-pas: It's at least not one of the existing ones; it takes a sprite pointer and explicitly skips `sub_08155128` afterwards, which no other macro has. I'm also not sure whether to put `SpriteInitMapObj` into `sprite.h`, as I haven't seen this behaviour in other files yet.
Thinking of it however, you could argue that it's a component of the other `SpriteInit*` macros, and I've now tested that substituting the assignment instructions (`tilesVram` to `unk8`) in `SpriteInit` and `SpriteInitNoPointer2` is indeed successful (weirdly enough `SpriteSomething` doesn't work). This can look a little bit nested, but I think it's cleaner.
I think it's easier to see what I mean if I put it in a separate commit, so it's in "Restructuring of `SpriteInit*` Macros" - and if you don't like it, I can just revert it.

src/pause_area_map.c
soiryl-pas: Yes, that should match this weird behaviour in the function.
As a caveat, in my decomp.me scratch with identical code (https://decomp.me/scratch/c50xb), it does however not fully replicate the control flow in assembly, especially with no outwards branch in the loop itself (like `0x17a` in target) - as it's not a `while(TRUE)`. This shouldn't really be a problem though, I think

jiangzhengwenjz: Great. I just wanted to double check.

PR #316, opened 2025-07-09
Split songs and label some of them

ultra-azu:
So far this is just the song splitting and the constants. If you approve what is here then I'll change the calls to m4aSongNumStart(and related) to use the constants in the header, either in this or in another PR, whatever you want.

### DIFF
--- a/include/constants/songs.h
+++ b/include/constants/songs.h
@@ -50,7 +50,7 @@
 #define MUS_BOSS_BRASS_UNUSED              0x035
 #define SE_KIRBY_JUMP                      0x064
 #define SE_KIRBY_INHALE                    0x065
-#define SE_KIRBY_STUFFED_CHEEKS_UNUSED            0x066
+#define SE_KIRBY_STUFFED_CHEEKS_UNUSED     0x066
 #define SE_KIRBY_STARSHOT_1                0x067
 #define SE_KIRBY_STARSHOT_2                0x068
 #define SE_08D54DE0                        0x069


jiangzhengwenjz:
> either in this or in another PR, whatever you want.

It doesn't really matter so maybe just do it here?

And I've a few comments:
1. Are all these `se` and `mus` files going to be converted back to midi? If not, I think they should be documented with macros in sound/MPlayDef.s. Considering the scale they are probably not going to be documented manually.
2. Idk if this can help but you can look into src/sound_test_debug.c for some names. There's no correspondence though.

Correct me if I said anything wrong for lack of knowledge in GBA audio.

ultra-azu:
> 1. Are all these se and mus files going to be converted back to midi? If not, I think they should be documented with macros in sound/MPlayDef.s. Considering the scale they are probably not going to be documented manually.

They are, eventually. Kurausukun told me splitting them was one of the steps for converting them to midi in the future. I could document them in MPlayDef.s since I won't be converting them to midi if you want to though (and I don't think it would be hard, I'm already using a python script to generate the linker.ld and the header after all).

> 2. Idk if this can help but you can look into src/sound_test_debug.c for some names. There's no correspondence though.

That's useful thx! I will rename them to match that more. For the ones that have a romanized Japanese name I should still name them in English though right?

jiangzhengwenjz:
> They are, eventually.

Then you don't have to document them.

> For the ones that have a romanized Japanese name I should still name them in English though right?
And I don't have to follow sound_test_debug too strictly right?

Right. The strings are only for reference.

ultra-azu:
In that case I renamed a few of them to match a bit better.

include/constants/songs.h
```diff
@@ -50,7 +50,7 @@
 #define MUS_BOSS_BRASS_UNUSED              0x035
 #define SE_KIRBY_JUMP                      0x064
 #define SE_KIRBY_INHALE                    0x065
-#define SE_KIRBY_INHALED_UNUSED            0x066
+#define SE_KIRBY_STUFFED_CHEEKS_UNUSED            0x066
```
jiangzhengwenjz: Renaming caused the value 0x066 to be misaligned with the others in the list.

jiangzhengwenjz (changes requested):

include/constants/songs.h
ultra-azu: Oops, done!

jiangzhengwenjz (approved):

PR #323, opened 2025-08-01
Decompilation of pause_world_map.s

Someone0123-pas:
Interestingly, almost all functions concern themselves with the area map, except for `sub_08127214` in between, which is only used during initialisation of the world map. If I had to guess, I can imagine all pause menu functions to have been in one big file. For this repo however, I'd rather like to have this separated to be cleaner, but I'll talk about that again when all pause stuff is decompiled.


src/pause_area_map.c
```diff
@@ -8,15 +9,52 @@
...
+extern const u8 gUnk_08363A90[8];
+extern const u8 gUnk_08363A98[8];
+// TODO when pause-menu file boundaries have been set:
+// data_22.s ends with gUnk_08363A98, perfect for defining data above
+
+extern const u32* gUnk_08D611C8[0xb];  // Backgrounds of areas (1-9) on the area map
+
+// Array of pointers to arrays with 0x200 size (except for last index, that's NULL)
+// extern const u16 (*gUnk_08D611F4[0xb])[0x100]; doesn't work?
+extern const u16* gUnk_08D611F4[0xb];
```
soiryl-pas: I tried to bake the size information into the type system to use sizeof and ARRAY_COUNT in `sub_0812752C`, but I'm not quite sure why the codegen is just a liiittle bit off (https://decomp.me/scratch/eFUL1).

I'd have thought this would then be an array of pointers pointing to arrays, but it looks like agbcc handles this a plain multidimensional array (which however also isn't quite right, the codegen for that would also be different). Is this an error of mine, or just agbcc?

src/pause_area_map.c
```diff
@@ -119,20 +168,410 @@ static inline void empty_screenbase_23(u32 arg0) {
...
+        areamapsprite1F0->unk0.x = areamapsprite120->unk0.x;
+        areamapsprite1F0->unk0.y = areamapsprite120->unk0.y;
+
+        x = areamapsprite1F0->unk0.x;
+        y = areamapsprite1F0->unk0.y;
+
+        xbounded = x + 0x1d <= 0x120;  // Needed for matching
+
+        // A coordinate bounds check that matches neither real screen pixels, nor with hblank and vblank included
+        r4 = x + 0x1d <= 0x120 && y + 0xf <= 0xbe;
```
soiryl-pas, line 275: This construction with the superfluous `xbounded` variable occurs in `sub_08127010` as well, have you perhaps seen this before? I find this coordinate checking to be a little bit weird, because from what I can see, this is slightly bigger than the pixel resolution of the screen, but slightly smaller than the resolution with hblank and vblank accounted.

src/pause_area_map.c
```diff
@@ -119,20 +168,410 @@ static inline void empty_screenbase_23(u32 arg0) {
...
+    unk6F4ptr = (u16*)(arg0->unk8 + arg0->unk0[1] * arg0->unkC);
+    unk6F4ptr++;
+    if (gMainFlags & MAIN_FLAG_BG_PALETTE_TRANSFORMATION_ENABLE) {
+        LoadBgPaletteWithTransformation(unk6F4ptr, arg0->unk0[0] / 2, (arg0->unk0[1] - 2) >> 1);
+    }
+    else {
+        // DmaCopy16(3, unk6F4ptr, gBgPalette + arg0->unk0[0]/2, arg0->unk0[1]-2);
+        // For matching, it must be a bitshift instead of normal divison like DmaCopy16 would do
+        DmaSet(3, unk6F4ptr, gBgPalette + arg0->unk0[0] / 2,
+               (DMA_ENABLE | DMA_START_NOW | DMA_16BIT | DMA_SRC_INC | DMA_DEST_INC) << 16 | (arg0->unk0[1] - 2) >> 1);
```
soiryl-pas, line 560: For some reason, this really must be a bitshift for both `LoadBgPaletteWithTransformation` and `DmaCopy16`, I haven't gotten normal division to work (https://decomp.me/scratch/hoM6N), and Compiler Explorer said the same thing.

data/data_1.s
```diff
@@ -2841,6 +2841,8 @@ gUnk_081E3990:: @ 081E3990
 gUnk_081E39AC:: @ 081E39AC
 	.incbin "baserom.gba", 0x1E39AC, 0x0000200
 
+@ Probably to be merged with gUnk_081E444D
+@ Background-Tileset of (not existing?) Area 0
```
jiangzhengwenjz: It's LZ77-compressed so they certainly should be merged

data/data_22.s
```diff
@@ -321,6 +321,7 @@ gUnk_08361DF4:: @ 08361DF4
 gUnk_08361E14:: @ 08361E14
 	.incbin "baserom.gba", 0x361E14, 0x0000194
 
+@ Probably to be merged with following, AreaMap UI tilemap
```
jiangzhengwenjz: It's LZ77-compressed so they certainly should be merged

src/pause_area_map.c
jiangzhengwenjz: > I tried to bake the size information into the type system to use sizeof and ARRAY_COUNT in `sub_0812752C`, but I'm not quite sure why the codegen is just a liiittle bit off (https://decomp.me/scratch/eFUL1).
> 
> I'd have thought this would then be an array of pointers pointing to arrays, but it looks like agbcc handles this a plain multidimensional array (which however also isn't quite right, the codegen for that would also be different). Is this an error of mine, or just agbcc?

As far as I know it's impossible in C.

src/pause_area_map.c
```diff
@@ -8,15 +9,52 @@
...
+extern const u8 gUnk_083636D8[3][4][4];
+extern const u8 gUnk_08363708[1][4][4];
+extern const u8 gUnk_08363718[3][4][4];
+extern const struct Unk_02021590 gUnk_08363898[][0xb];
+extern const u8 gUnk_08363A90[8];
+extern const u8 gUnk_08363A98[8];
+// TODO when pause-menu file boundaries have been set:
+// data_22.s ends with gUnk_08363A98, perfect for defining data above
+
+extern const u32* gUnk_08D611C8[0xb];  // Backgrounds of areas (1-9) on the area map
```
jiangzhengwenjz:
```suggestion
extern const u32* const gUnk_08D611C8[0xb];  // Backgrounds of areas (1-9) on the area map
```

src/pause_area_map.c
jiangzhengwenjz:
```suggestion
extern const u16* const gUnk_08D611F4[0xb];
```

src/pause_area_map.c
```diff
@@ -26,6 +64,8 @@ extern const u8 gUnk_08D61B20[0x1c];
 // index 4: R>>
 extern const u16 gUnk_08D61220[5][4];
 
+extern const u32* gUnk_08D61248[9];  // Table of pointers to area title tilesets
```
jiangzhengwenjz:
```suggestion
extern const u32* const gUnk_08D61248[9];  // Table of pointers to area title tilesets
```

src/pause_area_map.c
```diff
@@ -119,20 +168,410 @@ static inline void empty_screenbase_23(u32 arg0) {
...
+    }
+    return NULL;
+}
+
+static void sub_08126CEC(struct AreaMap* areamap) {
+    u32 r2;
+    for (r2 = 0; r2 < gUnk_0203AD44; r2++) {
+        const struct Unk_08361220* r3;
+        u16 currentRoom = gCurLevelInfo[r2].currentRoom;
+        if (9 <= gUnk_08D6CD0C[currentRoom]->unk46 && gUnk_08D6CD0C[currentRoom]->unk46 <= 10) {
```
jiangzhengwenjz:
```suggestion
        if (gUnk_08D6CD0C[currentRoom]->unk46 == 9 || gUnk_08D6CD0C[currentRoom]->unk46 == 10) {
```

src/pause_area_map.c
jiangzhengwenjz: > This construction with the superfluous `xbounded` variable occurs in `sub_08127010` as well, have you perhaps seen this before? I find this coordinate checking to be a little bit weird, because from what I can see, this is slightly bigger than the pixel resolution of the screen, but slightly smaller than the resolution with hblank and vblank accounted.

No, I've never seen before and it's hard to say anything until documenting the function

src/pause_area_map.c
```diff
@@ -119,20 +168,410 @@ static inline void empty_screenbase_23(u32 arg0) {
...
+
+static void sub_081270B8(s32 arg0, s32 arg1) {
+    if (arg0 < 1) return;
+    if (arg0 > 9) return;
+
+    LZ77UnCompVram(gUnk_08D61248[arg0 - 1], (void*)0x0600a800);
+    LZ77UnCompVram(gUnk_08361E14, (void*)0x0600b800);
+
+    if (arg1 != 2) {
+        empty_screenbase_23(1);
```
jiangzhengwenjz: Can we use pascal case for this? I know some inline functions are in snake case but it's only because I want to reuse `sub_xxx` in the name

src/pause_area_map.c
```diff
@@ -119,20 +168,410 @@ static inline void empty_screenbase_23(u32 arg0) {
...
+            if (gUnk_08350B30[gKirbys[r8].ability].animId) {
+                sub_0815604C(&areamapsprite1F0->unk0);
+            }
+        }
+    }
+}
+
+static void sub_08126F04(struct AreaMap* areamap) {
+    u8 r6;
+    CpuFill32(0, areamap->unk2C0, sizeof(areamap->unk2C0[0]) * 8);
```
jiangzhengwenjz, line 287: I wonder whether this is intentional or a mistake. From what I see `gUnk_08361A6E[areamap->unk6E0.unk6]` could exceed 8.

But anyway it probably doesn't matter since they are going to be initialized

src/pause_area_map.c
jiangzhengwenjz: > For some reason, this really must be a bitshift for both `LoadBgPaletteWithTransformation` and `DmaCopy16`, I haven't gotten normal division to work (https://decomp.me/scratch/hoM6N), and Compiler Explorer said the same thing.

I think it happened once before. Anyway it's fine as long as you manage to match it.

src/pause_area_map.c
```diff
@@ -119,20 +168,410 @@ static inline void empty_screenbase_23(u32 arg0) {
...
+            areamap->unk6E0.unkC = gUnk_083611F1[areamap->unk6E0.unk6][3] * 4;
+        }
+        areamap->unk6E0.unk4 = areamap->unk6E0.unk12;
+        break;
+    }
+}
+
+static void sub_08127760(struct AreaMap* areamap) {
+    if (areamap->unk48[areamap->unk6E0.unk6] == 2) {
+        RLUnCompVram(gUnk_0835EEF8 + gUnk_0835EECC[areamap->unk6E0.unk6], (void*)0x0600c000);
```
jiangzhengwenjz, line 525: According to GBATEK `gUnk_0835EEF8` should be 32-bit aligned (and every entry in `gUnk_0835EECC` is a multiple of 4).

src/pause_area_map.c
```diff
@@ -119,20 +168,410 @@ static inline void empty_screenbase_23(u32 arg0) {
...
+    areamap->unk0.prevScrollX = 0x7fff;
+    areamap->unk0.prevScrollY = 0x7fff;
+    areamap->unk0.paletteOffset = 0;
+    areamap->unk0.unk4 = 0x06004000;  // Charbase 1 -> BG0
+    areamap->unk0.unk2E = 0x18;
+    areamap->unk0.tilemapVram = 0x0600B000;  // Screenbase 22 -> BG0
+    LZ77UnCompVram((u32*)gUnk_08D611C8[unk6], (u32*)0x06004000);
+
+    if (gMainFlags & MAIN_FLAG_BG_PALETTE_TRANSFORMATION_ENABLE)
+        LoadBgPaletteWithTransformation((u16*)(gUnk_08D611F4[unk6] + 0x80), 0x80, 0x80);
```
jiangzhengwenjz: The u16* cast is not necessary

src/pause_area_map.c
```diff
@@ -119,20 +168,410 @@ static inline void empty_screenbase_23(u32 arg0) {
...
+    else {
+        DmaCopy16(3, (u16*)(gUnk_08D611F4[unk6] + 0x80), gBgPalette + 0x80, 0x100);
```
jiangzhengwenjz: The u16* cast is not necessary

jiangzhengwenjz (changes requested):

src/pause_area_map.c
soiryl-pas: Yep, changed it
I must admit, I named it ad-hoc bc normally I just append _helper to static inline functions I need for matching, but this one appeared in multiple functions, so I wanted a better name

src/pause_area_map.c
soiryl-pas: Yep, so this works, right? Or should I cast the first src parametre to a u32*
Oh, and on that topic, RL decompression into VRAM also wants the destination to be halfword aligned, is it generally better to cast dest to a u16* instead of void*, or does it rather not matter?

jiangzhengwenjz: Yes it works right now but can be easily broken when modding.
I mean you can have an attribute like this https://github.com/jiangzhengwenjz/katam/blob/master/include/palette.h#L8

> Oh, and on that topic, RL decompression into VRAM also wants the destination to be halfword aligned, is it generally better to cast dest to a u16* instead of void*, or does it rather not matter?

It doesn't really matter since the number (address) is explicitly defined by a literal

soiryl-pas: Ahh, this attribute is excellent, yes

jiangzhengwenjz (approved):

jiangzhengwenjz:
```
1413968 total bytes of code
1167608 bytes of code in src (82.5767%)
246360 bytes of code in asm (17.4233%)

41565 total symbols
6514 symbols documented (15.6718%)
35051 symbols undocumented (84.3282%)

14386872 total bytes of data
1222828 bytes of data in src (8.4996%)
13164044 bytes of data in data (91.5004%)
9918884 bytes of data in 29494 incbins (68.9440%)
```

PR #324, opened 2025-08-02
Decompilation of pause_help.s

Someone0123-pas:
Yeah, uh... I guess I was pretty motivated to get the pause menue stuff done, and most of these functions have been pretty nice to work with (with the exception of some perhaps rather weird, but necessary static inlines)
As mentioned, I included some stuff from code_08123950.s as well; some functions which hit breakpoints while navigating the pause menues.


src/code_08124BE0.c
```diff
@@ -104,6 +101,7 @@ void sub_08124BE0(void) {
             r8 = 2;
         }
         else {
+            // Could also be pause_help.c::sub_08128694_flags
```
jiangzhengwenjz: So does it work? I mean, we can put that static inline function in a header file if necessary

src/pause_help.c
```diff
@@ -0,0 +1,392 @@
...
+    gBgScrollRegs[1][0] = 0;
+    gBgScrollRegs[1][1] = 0;
+    gBgScrollRegs[2][0] = 0;
+    gBgScrollRegs[2][1] = 0;
+    gDispCnt = DISPCNT_MODE_0 | DISPCNT_OBJ_1D_MAP | DISPCNT_BG0_ON | DISPCNT_BG1_ON | DISPCNT_BG2_ON | DISPCNT_OBJ_ON;
+    gBgCntRegs[0] = BGCNT_PRIORITY(1) | BGCNT_CHARBASE(0) | BGCNT_16COLOR | BGCNT_SCREENBASE(7) | BGCNT_TXT256x256;
+    gBgCntRegs[1] = BGCNT_PRIORITY(0) | BGCNT_CHARBASE(1) | BGCNT_16COLOR | BGCNT_SCREENBASE(15) | BGCNT_TXT256x256;
+    gBgCntRegs[2] = BGCNT_PRIORITY(2) | BGCNT_CHARBASE(2) | BGCNT_16COLOR | BGCNT_SCREENBASE(23) | BGCNT_TXT256x256;
+
+    task = TaskCreate(sub_08124978, sizeof(struct HelpMenu), 0x1000, TASK_x0004, NULL);
```
jiangzhengwenjz:
```suggestion
    task = TaskCreate(sub_08124978, sizeof(struct HelpMenu), 0x1000, TASK_x0004 | TASK_USE_IWRAM, NULL);
```

src/pause_help.c
```diff
@@ -0,0 +1,392 @@
...
+
+static inline struct Task* TaskCreateWrapper(TaskMain taskMain, u16 structSize, u16 priority, u16 flags,
+                                             TaskDestructor taskDestructor) {
+    return TaskCreate(taskMain, structSize, priority, flags, taskDestructor);
+}
+
+// Selects which menu to show when pressing START
+// Called in sub_08039ED4 with function table gUnk_0834BD94
+void sub_0812424C(void) {
+    struct Task* task = TaskCreateWrapper(sub_08124BE0, 4, 0x0f00, TASK_x0004, NULL);
```
jiangzhengwenjz:
```suggestion
    struct Task* task = TaskCreateWrapper(sub_08124BE0, 4, 0x0f00, TASK_x0004 | TASK_USE_IWRAM, NULL);
```

jiangzhengwenjz: Is the allocated struct never used?

jiangzhengwenjz (changes requested):

src/code_08124BE0.c
soiryl-pas: Yes, it works! I'll put this static inline into pause_help.h then

src/pause_help.c
soiryl-pas: Not from what I've seen at least
I'm normally used to the TaskMain function using `TaskGetStructPtr(gCurTask)`, but that doesn't happen. As for the rest of TaskGetStructPtr calls in this file, they get used with different structs. If I had to guess, this could be debug stuff that was located in TaskCreateWrapper.

Also noticeably, while searching through code_08124BE0.c, `sub_08124F44` happens to create precisely the same Task, so instead of TaskCreateWrapper, this could actually be a non-static inline with this also marking an original TU boundary. I'll note that down for consideration when discussing the file boundaries later.

jiangzhengwenjz (approved):

PR #369, opened 2026-08-01
Match sub_080B819C, the last NONMATCHING function in flamer.c

jackmgaller:
Arms 3 and 4 write their four-coordinate setup out once per `flamer->base.flags & 1` arm rather than selecting the displacement first; cross-jumping merges the identical tails, and that is what lets `a << 16` be shared across arm 4's four endings.

Second commit moves the macro into `object.h` so `Macro_080B75D0` becomes a thin offset wrapper over it — no call-site changes, byte-identical.

Decompiled with AI assistance (Claude); each commit verified individually with a clean-tree `gmake compare`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jiangzhengwenjz (approved):
Thank you!!!

jiangzhengwenjz:
```
1413968 total bytes of code
1258084 bytes of code in src (88.9754%)
155884 bytes of code in asm (11.0246%)

41547 total symbols
6726 symbols documented (16.1889%)
34821 symbols undocumented (83.8111%)

14386873 total bytes of data
1263833 bytes of data in src (8.7846%)
13123040 bytes of data in data (91.2154%)
9868704 bytes of data in 29340 incbins (68.5952%)
```

PR #370, opened 2026-08-02
Match sub_08141300, dropping code_08138D64.c's hand-written bl

jackmgaller:
The `asm("ldr r0, [sp, #4]\n bl sub_0815604C\n")` was not needed — the plain call already produced every instruction correctly, and the only difference was two spill slots being exchanged. Reading `a1->unk9CA` into a local in `case 3` sheds one RTL insn, which moves gcse's expression hash table from 113 to 111 buckets and renumbers the two spilled address temporaries into the slots the target uses.

Six ROMs OK. This work was AI-assisted.


jiangzhengwenjz (approved):
I can never think of this as a human 🤦

PR #372, opened 2026-08-02
Name 25 Bonkers symbols, with a manifest

jackmgaller:
Renames 25 functions and 4 animation tables in `src/bonkers.c`. Every symbol is used only in `src/bonkers.c` and `include/bonkers.h`, so no other module is touched. Purely textual — all six ROMs match from a clean build.

`docs/naming/bonkers.md` gives the evidence for each decision individually, and records four symbols I deliberately left unnamed because I could not justify a name.

This work was AI-assisted.

### DIFF
--- a/include/bonkers.h
+++ b/include/bonkers.h
@@ -8,8 +8,8 @@
 void* CreateBonkers(struct Object*, u8);
 void* CreateBonkersNut(struct Object*, u8);
 
-void sub_080CF898(struct Object2*);
-void sub_080D1394(struct Object2*);
+void BonkersStartIdle(struct Object2*);
+void BonkersNutInit(struct Object2*);
 
 extern const struct AnimInfo gUnk_08356058[];
 extern const struct AnimInfo gUnk_083560B8[];
--- a/src/bonkers.c
+++ b/src/bonkers.c
@@ -5,33 +5,33 @@
 #include "inhalable_star.h"
 #include "code_0806F780.h"
 
-static void sub_080CF68C(struct Object2*);
-static void sub_080CF8F4(struct Object2*);
-static void sub_080CF960(struct Object2*);
-static void sub_080CFAF4(struct Object2*);
-static void sub_080CFC50(struct Object2*);
-static void sub_080CFF00(struct Object2*);
-static void sub_080D00AC(struct Object2*);
-static void sub_080D0180(struct Object2*);
+static void BonkersWaitForKirby(struct Object2*);
+static void BonkersStartWalk(struct Object2*);
+static void BonkersWalk(struct Object2*);
+static void BonkersRunWindup(struct Object2*);
+static void BonkersStartRunning(struct Object2*);
+static void BonkersRunning(struct Object2*);
+static void BonkersStartWallRecoil(struct Object2*);
+static void BonkersWallRecoil(struct Object2*);
 static void sub_080D02E0(struct Object2*);
 static void sub_080D0598(struct Object2*);
 static void sub_080D062C(struct Object2*);
-static void sub_080D078C(struct Object2*);
-static void sub_080D08D8(struct Object2*);
-static void sub_080D0AA0(struct Object2*);
-static void sub_080D0B9C(struct Object2*);
-static void sub_080D0D34(struct Object2*);
-static void sub_080D0DC0(struct Object2*);
-static void sub_080D109C(struct Object2*);
-static void sub_080D140C(struct Object2*);
-static void sub_080D1488(struct Object2*);
-static void sub_080D14AC(struct Object2*);
-static void sub_080D14C8(struct Object2*);
+static void BonkersHammerSwing(struct Object2*);
+static void BonkersHammerCombo(struct Object2*);
+static void BonkersJumpSlam(struct Object2*);
+static void BonkersJumpSlamRecover(struct Object2*);
+static void BonkersStartNutAttack(struct Object2*);
+static void BonkersNutAttack(struct Object2*);
+static void BonkersThrowNut(struct Object2*);
+static void BonkersNutFly(struct Object2*);
+static void BonkersStartWaitForKirby(struct Object2*);
+static void BonkersIdle(struct Object2*);
+static void BonkersStartRunWindup(struct Object2*);
 static void sub_080D14F8(struct Object2*);
-static void sub_080D1558(struct Object2*);
-static void sub_080D15B4(struct Object2*);
-static void sub_080D15F8(struct Object2*);
-static void sub_080D163C(struct Object2*);
+static void BonkersChooseAttack(struct Object2*);
+static void BonkersStartHammerSwing(struct Object2*);
+static void BonkersStartHammerCombo(struct Object2*);
+static void BonkersStartJumpSlam(struct Object2*);
 
 const struct AnimInfo gUnk_08356058[] = {
     { 0x31A,    1, 0 },
@@ -112,11 +112,11 @@ void *CreateBonkers(struct Object *arg0, u8 arg1) {
     obj->base.sprite.unk14 = 0x6c0;
     obj->unk9E = 0;
     obj->unk7C = sub_0809EF88;
-    sub_080D1488(obj);
+    BonkersStartWaitForKirby(obj);
     return obj;
 }
 
-static void sub_080CF68C(struct Object2 *arg0) {
+static void BonkersWaitForKirby(struct Object2 *arg0) {
     struct Kirby* kirby = sub_0803D368(&arg0->base);
     arg0->kirby3 = kirby;
     if (!(kirby->base.base.base.unkC & 0x8000)) {
@@ -130,7 +130,7 @@ static void sub_080CF68C(struct Object2 *arg0) {
             if (Macro_08039430_2(&arg0->kirby3->base.base.base, arg0)) {
                 Macro_081003EC(arg0, &arg0->kirby3->base.base.base);
                 arg0->base.flags &= ~0x200;
-                sub_080CF898(arg0);
+                BonkersStartIdle(arg0);
                 arg0->base.counter = 0x5a;
                 Macro_08100F18(arg0);
             }
@@ -138,8 +138,8 @@ static void sub_080CF68C(struct Object2 *arg0) {
     }
 }
 
-void sub_080CF898(struct Object2 *arg0) {
-    ObjectSetFunc(arg0, 0, sub_080D14AC);
+void BonkersStartIdle(struct Object2 *arg0) {
+    ObjectSetFunc(arg0, 0, BonkersIdle);
     arg0->base.xspeed = 0;
     arg0->base.yspeed = 0;
     arg0->base.counter = 0x1e;
@@ -149,8 +149,8 @@ void sub_080CF898(struct Object2 *arg0) {
     arg0->unk85 = 0;
 }
 
-static void sub_080CF8F4(struct Object2 *arg0) {
-    ObjectSetFunc(arg0, 2, sub_080CF960);
+static void BonkersStartWalk(struct Object2 *arg0) {
+    ObjectSetFunc(arg0, 2, BonkersWalk);
     if (arg0->subtype != 0) {
         arg0->base.xspeed = -0xc0;
     }
@@ -163,7 +163,7 @@ static void sub_080CF8F4(struct Object2 *arg0) {
     arg0->base.counter = Rand16() % 2;
 }
 
-static void sub_080CF960(struct Object2 *arg0) {
+static void BonkersWalk(struct Object2 *arg0) {
     arg0->base.flags |= 4;
     ObjXSomething(arg0);
     if (!(arg0->base.unk1 & 7)) {
@@ -177,7 +177,7 @@ static void sub_080CF960(struct Object2 *arg0) {
     }
     if (arg0->base.flags & 2) {
         if (arg0->base.counter == 0) {
-            sub_080D14C8(arg0);
+            BonkersStartRunWindup(arg0);
             return;
         }
         else {
@@ -189,7 +189,7 @@ static void sub_080CF960(struct Object2 *arg0) {
     }
 }
 
-static void sub_080CFAF4(struct Object2 *arg0) {
+static void BonkersRunWindup(struct Object2 *arg0) {
     arg0->base.flags |= 4;
     ObjXSomething(arg0);
     if (arg0->base.unk1 > 0xc) {
@@ -197,15 +197,15 @@ static void sub_080CFAF4(struct Object2 *arg0) {
     }
     
     if (arg0->base.flags & 2) {
-        sub_080CFC50(arg0);
+        BonkersStartRunning(arg0);
     }
     else if (arg0->base.unk62 & 2) {
         arg0->base.xspeed = 0;
     }
 }
 
-static void sub_080CFC50(struct Object2 *arg0) {
-    ObjectSetFunc(arg0, 4, sub_080CFF00);
+static void BonkersStartRunning(struct Object2 *arg0) {
+    ObjectSetFunc(arg0, 4, BonkersRunning);
     arg0->kirby3 = sub_0803D368(&arg0->base);
     if (arg0->base.x > arg0->kirby3->base.base.base.x) {
         arg0->base.flags |= 1;
@@ -276,7 +276,7 @@ static void sub_080CFC50(struct Object2 *arg0) {
     PlaySfx(&arg0->base, SE_MINIBOSS_RUN);
 }
 
-void sub_080CFF00(struct Object2 *arg0) {
+void BonkersRunning(struct Object2 *arg0) {
     ObjXSomething(arg0);
     arg0->base.flags |= 4;
     if (arg0->base.flags & 2) {
@@ -292,7 +292,7 @@ void sub_080CFF00(struct Object2 *arg0) {
     }
     if (arg0->base.x >= arg0->unkA0 * 0x100) {
         if (arg0->base.xspeed > 0) {
-            sub_080D1558(arg0);
+            BonkersChooseAttack(arg0);
             return;
         }
         else if (arg0->base.x > arg0->unkA0 * 0x100) {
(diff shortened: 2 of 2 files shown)


jiangzhengwenjz:
The manifest file is not supposed to be committed. You can upload it as an attachment

jackmgaller:
Naming manifest is attached here rather than committed to the tree.

**Downloadable file:** https://gist.github.com/jackmgaller/5adb1b0e60e08e4d6a1ce7e511129523

Inline copy below for convenience.

<details>
<summary><b>Naming manifest — Bonkers (29 renames)</b></summary>

# Naming manifest — Bonkers (`src/bonkers.c`, `include/bonkers.h`)

29 renames. Every symbol below is defined in `src/bonkers.c` and used nowhere
outside `src/bonkers.c` and `include/bonkers.h`, so no other module is affected.

Purely textual: all six ROMs match from a clean build.

## How to read the evidence

Bonkers is an `Object2` state machine. `ObjectSetFunc(obj, n, fn)`
(`src/object.c:2242`) does two things at once: it stores `n` into `obj->unk83`,
which selects the animation, and stores `fn` into `obj->unk78`, which becomes
the per-frame handler. So each transition function names both the animation and
the state it enters, and the pair `Start<X>` / `<X>` below is that split — the
transition that sets it up, and the handler that runs each frame.

Sound effects are the strongest single piece of evidence, because the constants
are already named in `include/constants/songs.h` by earlier contributors and are
independent of anything in this file.

Line numbers refer to upstream `src/bonkers.c` at `60e50b8`.

---

## Lifecycle

| old | new |
|---|---|
| `sub_080CF68C` | `BonkersWaitForKirby` |

Dormant handler installed by `CreateBonkers`. Each frame it re-acquires the
nearest Kirby (`sub_0803D368`), and does nothing until
`Macro_08039430_2(kirby, obj)` reports Kirby is close enough (:130); only then
does it clear the `0x200` flag, call `sub_080CF898` and start the fight. It never
moves the object and never plays a sound. Waiting for Kirby to approach is its
entire behaviour.

| old | new |
|---|---|
| `sub_080D1488` | `BonkersStartWaitForKirby` |

The transition into the above: `ObjectSetFunc(arg0, 0, sub_080CF68C)` and zeroed
x/y speed (:779). It is the last call in `CreateBonkers` (:115), so it is the
object's initial state.

| old | new |
|---|---|
| `sub_080CF898` | `BonkersStartIdle` |
| `sub_080D14AC` | `BonkersIdle` |

`sub_080CF898` sets animation 0, zeroes both speeds, and loads
`base.counter` with `0x1e` frames — halved to `0xf` when Bonkers is below half
health (`unk80 <= gUnk_08351530[1][...] >> 1`, :146). `sub_080D14AC` is a
two-line handler that decrements that counter and calls `sub_080CF8F4` when it
reaches zero (:785). A stationary countdown before acting is an idle state, and
the half-health branch shows the wait is deliberately shortened when Bonkers gets
more aggressive.

`sub_080CF898` is non-static and declared in `include/bonkers.h`; it has no C
caller outside this file, so it is reached from the object-type table, which is
still an `.incbin` blob upstream.

---

## Locomotion

| old | new |
|---|---|
| `sub_080CF8F4` | `BonkersStartWalk` |
| `sub_080CF960` | `BonkersWalk` |

`sub_080CF8F4` sets animation 2 and an x-speed of `0x80`, or `0xc0` for the
harder variant (`subtype != 0`), directed by the facing flag (:152). `sub_080CF960`
applies `ObjXSomething` each frame, re-faces Kirby every 8th frame, and exits to
`sub_080D14C8` when the animation completes. `0x80` is the slowest speed any
Bonkers state uses — compare `0x180`/`0x200` in the run below — so this is the
walk, not the run.

| old | new |
|---|---|
| `sub_080D14C8` | `BonkersStartRunWindup` |
| `sub_080CFAF4` | `BonkersRunWindup` |

`sub_080D14C8` sets animation 7 and a *reduced* speed of `0x80` (:791).
`sub_080CFAF4` zeroes x-speed once `base.unk1 > 0xc` and then calls
`sub_080CFC50` when the animation finishes (:192). A short decelerating state
that exists only to hand over to the run is a wind-up.

| old | new |
|---|---|
| `sub_080CFC50` | `BonkersStartRunning` |
| `sub_080CFF00` | `BonkersRunning` |

`sub_080CFC50` plays **`SE_MINIBOSS_RUN`** as its last statement (:276). It also
sets animation 4, raises x-speed to `0x180`/`0x200`, and picks a destination
`unkA0` — either a point `0x20` beyond Kirby, or, on the other branch, it walks
the collision map in `0x10` steps until it finds a wall or a tile with attribute
`0x200`, then stops short of it (:226-255). `sub_080CFF00` runs until
`base.x` passes `unkA0` and then calls `sub_080D1558`; if it hits a wall first
(`unk62 & 1`) it calls `sub_080D00AC` instead.

The sound effect alone fixes "run". Note this pair was `BonkersRun`/`BonkersRunning`
in the earlier branch; renamed to `BonkersStartRunning` here so it follows the
same `Start<X>`/`<X>` convention as every other pair in the file.

---

## Attack selection

| old | new |
|---|---|
| `sub_080D1558` | `BonkersChooseAttack` |

Pure dispatcher, no side effects of its own (:816). It picks between three
attacks and advances a 3-cycle counter:

- `unk85 == 2` → `sub_080D15F8` (hammer combo)
- else if Kirby is more than `0x1800` above Bonkers → `sub_080D163C` (jump slam)
- else → `sub_080D15B4` (hammer swing)
- then `if (++unk85 > 2) unk85 = 0`

Choosing among attacks, including one branch conditioned on Kirby's height, is
the whole function.

---

## Ground pound

| old | new |
|---|---|
| `sub_080D00AC` | `BonkersStartGroundPound` |
| `sub_080D0180` | `BonkersGroundPound` |

`sub_080D00AC` sets animation 5, launches Bonkers up and backwards
(`yspeed = 0x300`, `xspeed = -0x100` relative to facing), calls
`sub_0806FE64(2, ...)` and plays **`SE_BOSS_GROUND_POUND_ATTACK`** (:320).
`sub_080D0180` applies gravity while `unk83 == 5`; on touching the ground
(`unk62 & 4`) it plays **`SE_BOSS_GROUND_POUND_ATTACK`** again, calls
`sub_0806FE64(1, ...)` and switches to animation 6 (:332). Two firings of the
ground-pound sound, one on launch and one on impact, name this outright.

---

## Hammer attacks

| old | new |
|---|---|
| `sub_080D15B4` | `BonkersStartHammerSwing` |
| `sub_080D078C` | `BonkersHammerSwing` |

`sub_080D15B4` sets animation 11 and zeroes both speeds (:834). `sub_080D078C`
fires **exactly once**, at animation frame `0x12`: screen shake
(`sub_0806FE64(1, ...)`), an inhalable star spawned `0x28` ahead via
`sub_080A8C28` — the damage — and **`SE_BONKERS_HAMMER_ATTACK`** (:478-482).
One hammer sound, one hit: a single swing.

| old | new |
|---|---|
| `sub_080D15F8` | `BonkersStartHammerCombo` |
| `sub_080D08D8` | `BonkersHammerCombo` |

`sub_080D15F8` sets animation 17 (:846). `sub_080D08D8` plays
**`SE_BONKERS_HAMMER_ATTACK`** at **three** frames — `0x12`, `0x26` and `0x3a`
(:512-521) — where only the first also calls `sub_080A8C28`. Three impacts in one
state, against the single impact of `sub_080D078C`, is the distinction being
named.

These two were `BonkersDoubleHammerSwing` / `BonkersStartDoubleHammerSwing` in
the earlier branch. **That was wrong** — there are three impact frames, not two —
so "Combo" replaces "Double", which is what the code actually supports.

---

## Jump slams

| old | new |
|---|---|
| `sub_080D163C` | `BonkersStartJumpSlam` |
| `sub_080D0AA0` | `BonkersJumpSlam` |
| `sub_080D0B9C` | `BonkersJumpSlamRecover` |

`sub_080D163C` sets animation 12 and zeroes both speeds (:858).

`sub_080D0AA0` is the damaging jump: at `unk1 == 8` it sets `yspeed = 0x4a0`, the
largest upward impulse in the file; on landing it calls `sub_0806FE64(1, ...)`,
spawns the damaging star via `sub_080A8C28(arg0, 0x28, 8)`, plays
**`SE_BOSS_GROUND_POUND_ATTACK`**, moves to animation `0xe` and hands over to
`sub_080D0B9C` (:557-564).

`sub_080D0B9C` is the follow-up: a smaller hop (`yspeed = 0x280`), then animation
`0xf` → `0x10` with **`SE_BONKERS_JUMP`**, and after `0xf` more frames it picks
the next attack (:568-614). It contains **no `sub_080A8C28` call**, so it deals
no damage — it is recovery, not an attack.

The earlier branch had these as `BonkersJumpSlamWindup` / `BonkersJumpSlam`,
i.e. shifted by one. **The damage call and the ground-pound sound are both in
`sub_080D0AA0`**, so that is the slam, and `sub_080D0B9C` is what follows it.

| old | new |
|---|---|
| `sub_080D0D34` | `BonkersStartBigJumpSlam` |
| `sub_080D0DC0` | `BonkersBigJumpSlam` |

`sub_080D0D34` sets animation `0x12` and, for `subtype1 == 0`, a travelling arc
(`xspeed = -0x180`, `yspeed = 0x280`) (:617). `sub_080D0DC0` applies
`ObjXSomething` so it moves horizontally through the jump — the plain jump slam
above does not — and on landing plays **`SE_BONKERS_JUMP`** with two
`sub_08089864` shockwaves (:646-656). At frame `0x44` it calls `sub_080D109C`
to throw a nut (:659). A travelling jump with a wider impact and a projectile is
distinguished from the stationary one by "Big".

---

## Thrown nuts

| old | new |
|---|---|
| `sub_080D109C` | `BonkersThrowNut` |

Spawns one object `0x18` ahead of Bonkers, choosing `OBJ_BONKERS_NUT_SMALL` or
`OBJ_BONKERS_NUT_LARGE` by `RandLessThan3()`, and parents it to Bonkers
(:690-709). Both type constants were already named upstream in
`include/constants/object_types.h`, so the object being thrown is not a guess.
Named `BonkersThrowNut` rather than the earlier `BonkersThrowObject` because the
two `OBJ_BONKERS_NUT_*` constants are the only things it can spawn.

| old | new |
|---|---|
| `sub_080D1394` | `BonkersNutInit` |
| `sub_080D140C` | `BonkersNutFly` |

`sub_080D1394` is called at the end of `CreateBonkersNut` (:728). It selects
animation 1 for `OBJ_BONKERS_NUT_LARGE` and 0 otherwise, then sets the launch arc
`xspeed = 0x180`, `yspeed = 0x1e0`, halving x-speed for the large nut (:733).
`sub_080D140C` is the nut's per-frame handler: on the first ground contact it
bounces once with reduced speed, and on the second contact — or on hitting a wall
— it calls `sub_0809DA30` and sets the `0x1000` despawn flag (:753). Like
`BonkersStartIdle`, `sub_080D1394` is exported in the header with no C caller, so
it is reached from the object-type table.

---

## Animation tables

| old | new |
|---|---|
| `gUnk_08356058` | `gBonkersAnimInfo` |
| `gUnk_083560B8` | `gBonkersAnimInfo2` |
| `gUnk_083560C0` | `gBonkersAnimInfo3` |
| `gUnk_083560FC` | `gBonkersAnimInfo4` |

All four are `const struct AnimInfo[]` defined in `src/bonkers.c` and declared in
`include/bonkers.h`; every entry references sprite `0x31A`, which no other file
uses.

`gUnk_08356058` is the main table: 24 entries whose animation ids run `0x00`
through `0x15` in order. The state functions above pass `0`–`0x13` as the second
argument of `ObjectSetFunc`, which stores it in `unk83`, so this table is indexed
by exactly the range those states produce.

The other three are 2, 15 and 3 entries. They have **no reference anywhere in C
or in `data/`** — they are reached from the object-type table, which is still an
`.incbin` blob upstream, so I cannot say from the code what each one drives. The
numeric suffixes are therefore only disambiguation; the part being asserted is
"these are Bonkers' animation tables", which follows from the file they live in
and the sprite id they share. If you would rather these stayed `gUnk_`, say so
and I will drop them — the other 25 renames do not depend on them.

---

## Deliberately not renamed

Four symbols in this file are left as `sub_XXXXXXXX`. Both pairs are a
hammer-slam transition plus its handler, and the two handlers share animations
8, 9 and 10, so the animation ids do not separate them:

| symbol | what is clear | why I did not name it |
|---|---|---|
| `sub_080D14F8` | sets animation 8, enters `sub_080D02E0` | see below |
| `sub_080D02E0` | rises for 7 frames, then leaps toward Kirby (`xspeed` set toward him, `yspeed = 0x600`), lands with two `sub_08089864` shockwaves and `SE_BOSS_GROUND_POUND_ATTACK` | a travelling slam |
| `sub_080D0598` | sets animation 8, enters `sub_080D062C`, and seeds `base.counter` with 1 or 3 | see below |
| `sub_080D062C` | same rise, but drops straight down with **no** x-speed, and repeats `base.counter` times before ending | a stationary, repeating slam |

The behavioural difference is real and I have described it, but I could not
settle on names that state it without inventing vocabulary the game does not
use, and a numbered pair (`...Slam` / `...Slam2`) would assert an ordering the
code does not support. Left unnamed pending a better reading.

</details>

include/bonkers.h
```diff
@@ -8,12 +8,12 @@
...
+void BonkersNutInit(struct Object2*);
 
-extern const struct AnimInfo gUnk_08356058[];
-extern const struct AnimInfo gUnk_083560B8[];
-extern const struct AnimInfo gUnk_083560C0[];
-extern const struct AnimInfo gUnk_083560FC[];
+extern const struct AnimInfo gBonkersAnimInfo[];
+extern const struct AnimInfo gBonkersAnimInfo2[];
+extern const struct AnimInfo gBonkersAnimInfo3[];
+extern const struct AnimInfo gBonkersAnimInfo4[];
```
jiangzhengwenjz: They are not wrong but provide zero knowledge.

src/bonkers.c
```diff
@@ -637,7 +637,7 @@ static void sub_080D0D34(struct Object2 *arg0) {
     arg0->base.flags |= 0x40;
 }
 
-static void sub_080D0DC0(struct Object2 *arg0) {
+static void BonkersBigJumpSlam(struct Object2 *arg0) {
```
jiangzhengwenjz: I guess the function always inflicts nut throw? If so, it's not reflected in the name

src/bonkers.c
```diff
@@ -5,35 +5,35 @@
...
-static void sub_080D00AC(struct Object2*);
-static void sub_080D0180(struct Object2*);
+static void BonkersWaitForKirby(struct Object2*);
+static void BonkersStartWalk(struct Object2*);
+static void BonkersWalk(struct Object2*);
+static void BonkersRunWindup(struct Object2*);
+static void BonkersStartRunning(struct Object2*);
+static void BonkersRunning(struct Object2*);
+static void BonkersStartGroundPound(struct Object2*);
+static void BonkersGroundPound(struct Object2*);
```
jiangzhengwenjz: I don't think they make any sense. And you probably need to modify `SE_BOSS_GROUND_POUND_ATTACK` since it only partially documents the usage for the SE. Please consider this carefully.

jiangzhengwenjz (changes requested):

include/bonkers.h
jackmgaller: Agreed — reverted, all four are back to `gUnk_*` and this PR no longer touches them.

You're right that the suffixes carry nothing. Checking afterwards, the convention is also universal: **158 of 158 `AnimInfo` externs in `include/` are `gUnk_*`**, no exceptions, so naming four in one file would have made Bonkers the odd one out. And three of these have no reference anywhere in C or `data/` — they're reached by address from `gUnk_08D61048[]`/`gUnk_08D610B4[]`, still `.incbin` upstream — so there's no use site to name them after in the first place.

src/bonkers.c
jackmgaller: Yes, it always throws. Renamed to `BonkersNutAttack` (and `sub_080D0D34` → `BonkersStartNutAttack`).

Two things make the throw the invariant and the jump the incidental part:

- Animation `0x13` runs `if (arg0->base.unk1 == 0x44) BonkersThrowNut(arg0);` — unconditional at that frame.
- `sub_080D0D34` **skips the jump entirely** when `object->subtype1 != 0`: it assigns `unk83 = 0x13` directly and never leaves the ground. So one subtype of this state is a throw with no jump at all.

Also worth noting the landing plays `SE_BONKERS_JUMP`, not any impact sound, which is further reason "Slam" was wrong.

src/bonkers.c
jackmgaller: You're right on both counts, and the ground-pound pair was worse than "doesn't make sense" — it was actively wrong. Renamed to `BonkersStartWallRecoil` / `BonkersWallRecoil`.

**The entry point settles it.** The only caller is the last line of `BonkersRunning`:

```c
if (arg0->base.unk62 & 1)
    BonkersStartWallRecoil(arg0);
```

and `code_080023A4.c:2448` sets `unk62 |= (flags & 1) != 0 ? 1 : 2` immediately after clamping `x`, so bit 0 means *hit a wall in the direction of travel*. The state is reachable only by running into a wall. It then sends Bonkers **backwards** (`xspeed = -0x100` relative to facing, `yspeed = 0x300`), lands, shakes the screen, and jitters `yspeed` between ±0x100 on a four-frame cycle for ~40 frames before choosing a new attack. **No `sub_080A8C28` call anywhere in either function**, unlike every hammer state. That's a crash and a daze, not an attack.

**On `SE_BOSS_GROUND_POUND_ATTACK`:** agreed, and I've stopped using it as evidence. It's played by eight bosses at eighteen sites — `phan_phan.c` ×4, `bonkers.c` ×5, `mr_frosty.c` ×3, `mega_titan.c` ×2, `boxy.c` ×2, `dark_meta_knight.c`, `object.c` — and every one is a heavy body landing. It names an impact, not an attack.

Correcting the constant touches seven other files, which is exactly the cross-system naming you asked to keep out of per-system PRs, so I left it alone here. Happy to send it as its own small PR — would you prefer the constant renamed (something like `SE_BOSS_HEAVY_LANDING`), or just its comment corrected to list the real callers? Your call.

The remaining names in this hunk I'd defend: `SE_MINIBOSS_RUN` is played by `BonkersStartRunning`, `BonkersRunWindup` zeroes `xspeed` after `unk1 > 0xC` and transitions on animation end, and `BonkersWalk` re-faces Kirby every 8 frames. Say the word if any of those look shaky too.

jackmgaller:
Revision 2 pushed, addressing all three comments. 29 renames → **25**, all functions.

- **Animation tables** — withdrawn, all four back to `gUnk_*`.
- **`BonkersBigJumpSlam`** → `BonkersNutAttack` — the throw is unconditional and one subtype skips the jump entirely.
- **`BonkersGroundPound`** → `BonkersWallRecoil` — reachable only via `unk62 & 1` (wall hit while running), moves backwards, no damage call.

All three were the same mistake: naming from `SE_BOSS_GROUND_POUND_ATTACK` or from a table's position rather than from the code. The manifest now says so explicitly and no longer treats that constant as evidence.

Open question for you in the third thread: whether you want `SE_BOSS_GROUND_POUND_ATTACK` renamed or just its comment corrected — as a separate PR either way, since it reaches seven other files.

**Updated manifest:** https://gist.github.com/jackmgaller/5adb1b0e60e08e4d6a1ce7e511129523

Six ROMs verified from a clean build at `11ee931`.

jiangzhengwenjz (approved):

jiangzhengwenjz:
Thanks for the contribution.

PR #373, opened 2026-08-02
Decompile sword_knight, vert_sliding_door and eight_dir_cannon (25 functions)

jackmgaller:
All 25 functions match — including `EightDirCannonRotate`, using the body from review, so nothing here ships as NONMATCHING.

Removes 3,489 lines of asm. Rebased onto `c859c3b`; all six ROMs verified from a clean build.

AI-assisted (Claude).


src/eight_dir_cannon.c
```diff
@@ -0,0 +1,235 @@
...
+                    cannon2->unkB4 = 0;
+                    cannon->unkB7 |= 2;
+                }
+                cannon->obj2.unk83 = cannon2->unkB4;
+                cannon2->unkB5 = 5;
+            }
+        }
+    }
+}
+#endif
```
jiangzhengwenjz:
```suggestion
static void EightDirCannonRotate(struct EightDirCannon *cannon) {
    struct EightDirCannon *cannon2 = cannon;
    u8 i;

    if (cannon->unkB7 & 1) {
        if (!(cannon->obj2.object->unk22 & 1)) {
            if (cannon2->unkBC->unk11A & 1) {
                for (i = 0; i < gUnk_0203AD44; i++) {
                    if ((cannon2->unkBA >> i) & 1)
                        sub_08051F70(&gKirbys[i], cannon2->unkB4);
                }
                cannon->obj2.unk78 = EightDirCannonRecoil;
                cannon2->unkB7 &= ~8;
                sub_081222AC(&cannon->obj2.base, cannon2->unkB4);
                sub_0806FE64(3, &cannon->obj2.base);
            }
        }
        else if (cannon2->unkB7 & 2 && cannon2->unkB4 == cannon->obj2.object->unk14) {
            for (i = 0; i < gUnk_0203AD44; i++) {
                if ((cannon2->unkBA >> i) & 1)
                    sub_08051F70(&gKirbys[i], cannon2->unkB4);
            }
            cannon->obj2.unk78 = EightDirCannonRecoil;
            cannon2->unkB7 &= ~8;
            sub_081222AC(&cannon->obj2.base, cannon2->unkB4);
            sub_0806FE64(3, &cannon->obj2.base);
        }
        if (cannon2->unkB7 & 8) {
            cannon2->unkB5--;
            if (cannon2->unkB5 == 0xFF) {
                cannon2->unkB4++;
                if (cannon2->unkB4 > 7) {
                    cannon2->unkB4 = 0;
                    cannon2->unkB7 |= 2;
                }
                cannon->obj2.unk83 = cannon2->unkB4;
                cannon2->unkB5 = 5;
            }
        }
    }
}
```

src/eight_dir_cannon.c
```diff
@@ -0,0 +1,235 @@
...
+static void EightDirCannonIdle(struct EightDirCannon *);
+static void EightDirCannonRotate(struct EightDirCannon *);
+static bool32 EightDirCannonUpdateRecoil(struct EightDirCannon *);
+static void EightDirCannonRecoil(struct EightDirCannon *);
+
+extern void sub_081222AC(struct ObjectBase *, u8);
+extern bool8 sub_08051C40(struct Kirby *);
+extern void sub_08051F70(struct Kirby *, u8);
+
+void *Create8DirCannon(struct Object *arg0, u8 arg1) {
```
jiangzhengwenjz: `Eight` is used everywhere else

src/sword_knight.c
```diff
@@ -0,0 +1,432 @@
...
+static void sub_080B2E6C(struct Object2 *);
+static void sub_080B2F3C(struct Object2 *);
+static void sub_080B2FE4(struct Object2 *);
+static void sub_080B31F0(struct Object2 *);
+static void sub_080B3420(struct Object2 *);
+static void sub_080B351C(struct Object2 *);
+static void sub_080B3694(void);
+static void sub_080B37DC(struct Object2 *);
+static void sub_080B38B0(void);
+static void sub_080B3A18(struct Object2 *);
```
jiangzhengwenjz: Should not be `static` since it's referenced in `gUnk_08351648`

src/sword_knight.c
```diff
@@ -0,0 +1,432 @@
...
+            obj->base.xspeed = -obj->base.xspeed;
+        if (sub_0809D998(obj))
+            obj->base.xspeed = 0;
+    }
+    if (obj->object->subtype1 <= 1) {
+        if (obj->base.flags & 2)
+            obj->kirby3 = sub_0803D368(&obj->base);
+        if (obj->kirby3->base.base.base.x - obj->base.x >= 0
+                ? obj->kirby3->base.base.base.x - obj->base.x <= 0x3FFF
+                : obj->base.x - obj->kirby3->base.base.base.x <= 0x3FFF) {
```
jiangzhengwenjz: use `abs`

src/sword_knight.c
```diff
@@ -0,0 +1,432 @@
...
+    obj->base.flags &= ~8;
+    PlaySfx(&obj->base, SE_08D5859C);
+}
+
+static void sub_080B351C(struct Object2 *obj) {
+    struct Task *t = TaskCreate(sub_080B3694, sizeof(struct ObjectBase), 0x3500, TASK_USE_EWRAM, NULL);
+    struct ObjectBase *tmp = TaskGetStructPtr(t), *objBase;
+    u32 flags;
+
+    if (tmp) objBase = tmp; // see also: sub_080BF914
```
jiangzhengwenjz:
```suggestion
```

src/sword_knight.c
```diff
@@ -0,0 +1,432 @@
...
+                else {
+                    objBase->x = x + 0x1A00;
+                    objBase->flags &= ~1;
+                }
+                if (!(objBase->flags & 0x200))
+                    SetPointerSomething(objBase);
+            }
+        }
+    }
+}
```
jiangzhengwenjz, line 317:
```suggestion
static void sub_080B3694(void) {
    struct ObjectBase *tmp = TaskGetStructPtr(gCurTask), *objBase = tmp;
    struct Object2 *obj = objBase->parent;

    if (objBase->flags & 0x1000)
        TaskDestroy(gCurTask);
    else if (obj->base.flags & 0x1000)
        objBase->flags |= 0x1000;
    else {
        objBase->unk56 = obj->base.unk56;
        objBase->x = obj->base.x;
        objBase->y = obj->base.y;
        if (!Macro_0810B1F4(&obj->base)) {
            if (++objBase->counter > 3) {
                objBase->flags |= 0x1000;
            }
            else {
                if (obj->base.flags & 1) {
                    objBase->x -= 0x1A00;
                    objBase->flags |= 1;
                }
                else {
                    objBase->x += 0x1A00;
                    objBase->flags &= ~1;
                }
                if (!(objBase->flags & 0x200))
                    SetPointerSomething(objBase);
            }
        }
    }
}
```

src/sword_knight.c
```diff
@@ -0,0 +1,432 @@
...
+    objBase = tmp;
+    sub_0803E380(objBase);
+    objBase->unk0 = 2;
+    objBase->x = obj->base.x;
+    objBase->y = obj->base.y;
+    objBase->parent = obj;
+    objBase->counter = 0;
+    objBase->roomId = obj->base.roomId;
+    objBase->unk56 = obj->base.unk56;
+    if ((u16)Macro_0810B1F4(objBase))
```
jiangzhengwenjz:
```suggestion
    if (Macro_0810B1F4(objBase))
```

src/sword_knight.c
```diff
@@ -0,0 +1,432 @@
...
```
jiangzhengwenjz:
```suggestion
    if (Macro_0810B1F4(objBase))
```

src/vert_sliding_door.c
```diff
@@ -0,0 +1,106 @@
...
+
+void VertSlidingDoorSetFunc(struct VertSlidingDoor *arg0) {
+    struct VertSlidingDoor *door = arg0;
+
+    door->obj2.base.counter = 0;
+    door->obj2.unk78 = VertSlidingDoorIdle;
+    door->unkB4 = 0;
+    door->unkB6 = 0;
+    door->unkB7 = 0;
+}
```
jiangzhengwenjz:
```suggestion
void VertSlidingDoorSetFunc(struct VertSlidingDoor *door) {
    door->obj2.base.counter = 0;
    door->obj2.unk78 = VertSlidingDoorIdle;
    door->unkB4 = 0;
    door->unkB6 = 0;
    door->unkB7 = 0;
}
```

src/sword_knight.c
```diff
@@ -0,0 +1,432 @@
...
+    else
+        obj->base.flags &= ~1;
+    obj->base.flags &= ~8;
+    PlaySfx(&obj->base, SE_08D5859C);
+}
+
+static void sub_080B351C(struct Object2 *obj) {
+    struct Task *t = TaskCreate(sub_080B3694, sizeof(struct ObjectBase), 0x3500, TASK_USE_EWRAM, NULL);
+    struct ObjectBase *tmp = TaskGetStructPtr(t), *objBase;
+    u32 flags;
```
jiangzhengwenjz, line 258: Possible to eliminate the var?

src/sword_knight.c
```diff
@@ -0,0 +1,432 @@
...
+                    SetPointerSomething(objBase);
+            }
+        }
+    }
+}
+
+static void sub_080B37DC(struct Object2 *obj) {
+    struct Task *t = TaskCreate(sub_080B38B0, sizeof(struct ObjectBase), 0x3500, TASK_USE_EWRAM, NULL);
+    struct ObjectBase *tmp = TaskGetStructPtr(t), *objBase;
+    u32 flags;
```
jiangzhengwenjz, line 322: Possible to eliminate the var?

jiangzhengwenjz (changes requested):

src/eight_dir_cannon.c
jackmgaller: Applied — and it **matches**. Six ROMs OK, and I confirmed with a negative control (changed one constant, `katam.gba: FAILED`).

So this PR no longer ships anything NONMATCHING and there's no `asm/nonmatching` entry. Thank you — I'd concluded the `ip` placement wasn't reachable from C, and it plainly is; routing the accesses through `cannon2` rather than `cannon` was the part I had backwards.

src/eight_dir_cannon.c
jackmgaller: Done — `CreateEightDirCannon`. The reference in `data/object.s:149` (`gSpawnFuncTable1`) is updated in the same commit, since it's the only other use.

src/sword_knight.c
jackmgaller: Fixed. Confirmed against the ROM before changing it — `gUnk_08351648[26]` (0x18 stride) holds `unk10 = 0x080B3A19` and `unk14 = 0x08353FA8`, so both the init function and the anim table are reached only through that entry and neither has a caller in the file. Declared in `include/sword_knight.h` alongside `CreateSwordKnight`, matching `VertSlidingDoorSetFunc` / `EightDirCannonSetFunc` in this same PR.

src/sword_knight.c
jackmgaller: Applied, and it matches. I'd assumed the macro would emit `negs` where the target recomputes the subtraction in the other order, so I hand-wrote the ternary — wrong, agbcc canonicalises it. `batty.c:151` uses the same idiom and I should have followed it.

src/sword_knight.c
jackmgaller: Applied — the cast wasn't needed. I'd copied it from `box_boxer.c:707`, which carries a `// the cast is required for matching` comment; it is not required here.

src/sword_knight.c
jackmgaller: Applied here too, same as the other one.

src/sword_knight.c
jackmgaller: Applied, and it matches — **once the two `(u16)` casts came off**. On its own this rewrite didn't match, which is why I'd added the `s32 x` local: the target keeps `obj->base.x` in one register across both arms, and with the casts present my build reloaded it each time. Dropping the casts changed the allocation enough that `objBase->x -= 0x1A00` gets there directly. Retesting a rejected variant after each accepted change is clearly the habit I was missing.

src/vert_sliding_door.c
jackmgaller: Applied — matches.

src/sword_knight.c
jackmgaller: This one I had to keep — dropping it breaks the ROM. I tested two levels:

1. delete just this line, keeping both declarators → `katam.gba: FAILED`
2. collapse to `struct ObjectBase *objBase = TaskGetStructPtr(t);` → `katam.gba: FAILED`

Retested both again at the end, after the `abs`, cast and `sub_080B3694` changes had landed, in case the interaction freed it up the way it did for `s32 x` — still fails.

What it buys: the target computes the `TaskGetStructPtr` ternary into `r0` and *then* copies it (`adds r4, r0, #0`), leaving `r0` live as the argument to `sub_0803E380`. Without the redundant assignment agbcc computes straight into `r4` and copies back to `r0`, which is a different instruction.

It's ugly and I'd rather not have it. It is the same idiom as `box_boxer.c:696`, which carries `// see also: sub_080BF914` for the same reason, so at least it's consistent with what's already in the tree — but if you'd prefer it spelled some other way I'm happy to try a specific suggestion.

src/sword_knight.c
jackmgaller: `flags` I also couldn't remove. Writing it as `objBase->flags = 0x50500400; ... objBase->flags |= 1;` and dropping the local → `katam.gba: FAILED` (retested at the end, after the other changes). The target stores the constant, keeps it in a register across the x adjustment, ORs bit 0 in the register and stores again; going through the field directly gives a load-modify-store instead.

`tmp`/`objBase` on the line above is load-bearing too — details in the thread below.

src/sword_knight.c
jackmgaller: Same as `sub_080B351C` — removing `flags` here fails as well (`0x42110400` case). Both retested after the other review changes landed.

jackmgaller:
Rebased onto `c859c3b` and revised. **9 of 11 comments applied; 2 I couldn't take, with measurements in their threads.**

The headline: **your `EightDirCannonRotate` body matches.** 25/25 now, and this PR ships no NONMATCHING at all. I'd written that one off as unreachable from C — routing the accesses through `cannon2` instead of `cannon` was the thing I had backwards.

Applied: the `EightDirCannonRotate` body · `CreateEightDirCannon` (+ its `data/object.s` reference) · `sub_080B3A18` un-`static` · `abs()` · both `(u16)` casts dropped · the `sub_080B3694` rewrite · `VertSlidingDoorSetFunc`.

Not applied, both verified to break `katam.gba` and both retested at the end in case the other changes freed them up: deleting `if (tmp) objBase = tmp;`, and eliminating `u32 flags` in the two spawn functions.

Worth noting one interaction: **your `sub_080B3694` rewrite only matches once the `(u16)` casts are removed.** On its own it didn't — that's why the `s32 x` local was there. Retesting rejected variants after each accepted change is the habit I was missing.

Thanks also for the `SE_BOSS_GROUND_POUND_ATTACK` TODO in `c859c3b` — that answers the question I left on #372.

src/sword_knight.c
jiangzhengwenjz: I tested locally and deleting the line is not going to break anything.

src/sword_knight.c
jiangzhengwenjz: I certainly know that removing `flags` itself is not going to help. I meant if it's possible to mutate the current version of this function to match without having the variable.

src/sword_knight.c
jiangzhengwenjz: Same comment as that on `sub_080B351C`

src/sword_knight.c
jackmgaller: You're right, and I had a sloppy test — sorry. Removed in `sub_080B351C`; pushed.

My earlier check deleted the line from **both** spawn functions at once and I reported the combined result. Splitting them:

| function | with the line | without |
|---|---|---|
| `sub_080B351C` | 169 insns | **byte-identical** |
| `sub_080B37DC` | 101 insns | 3 instructions differ |

So it does nothing in the one you tested. In `sub_080B37DC` it still costs:

```
-adds r0, r0, r1          +adds r4, r0, r1
-adds r0, r1, r0          +adds r4, r1, r0
-adds r4, r0, #0          +adds r0, r4, #0
```

The target computes the `TaskGetStructPtr` ternary into `r0` and copies to `r4`, leaving `r0` live as the argument to `ClearObjectBase`; without the redundant assignment agbcc computes into `r4` and copies back. The difference between the two functions seems to be that `sub_080B351C` ends with a `PlaySfx` call and `sub_080B37DC` doesn't, so the register pressure differs.

Happy to drop it there too if you'd rather take the 3-instruction regression, but as it stands it's load-bearing in exactly one of the two.

src/sword_knight.c
jackmgaller: Understood — I'd answered the wrong question. I swept five restructurings that keep the same semantics with no local, comparing per-function objdump against the matching build:

| formulation | result |
|---|---|
| `objBase->flags = C;` … `objBase->flags \|= 1;` | 169 vs 167 |
| both arms store an explicit constant (`C`, `C\|1`) | 170 vs 167 |
| store `C`, then `objBase->flags = C \| (obj->base.flags & 1);` at the join | 172 vs 167 |
| fold the OR into the initial store, adjust x separately | 169 vs 167 |
| `\|= 1` in one arm, `&= ~1` in the other | 169 vs 167 |

The obstruction looks structural rather than cosmetic. The target keeps the constant in a register across the branch and stores **once** at the join:

```
orrs r2, r3          @ set bit 0 in the register
b    <join>
<join>:
str  r2, [r4, #8]    @ single store
```

Every no-local form has to commit to memory inside the arm, which costs a move plus a store and loses the shared tail:

```
adds r0, r3, #0
orrs r0, r2
str  r0, [r4, #8]    @ store inside the arm
b    <join>
```

That's the +2. So the local isn't there to hold a value — it's there to keep the store out of the branch.

If you have a particular shape in mind I'll test it directly; the sweep harness is set up now, so it's a couple of minutes per candidate.

src/sword_knight.c
jackmgaller: Same outcome for `sub_080B37DC` — its `flags` block is the `0x42110400` variant of the same shape, and removing the local costs the same store-inside-the-arm. Measurements in the `sub_080B351C` thread.

jackmgaller:
Pushed `135f235`. `if (tmp) objBase = tmp;` is gone from `sub_080B351C` — you were right, and my earlier "it fails" was a bad test: I'd deleted it from both spawn functions and reported the combined result. It's a no-op in that one and costs 3 instructions in `sub_080B37DC`, so it now appears only there.

On `flags`: swept five no-local restructurings, all 2–5 instructions worse. Detail in the threads — short version is the local keeps the store out of the branch arm, not a value.

Six ROMs verified from a clean build.

jiangzhengwenjz (approved):

jiangzhengwenjz:
```
1413968 total bytes of code
1264880 bytes of code in src (89.4561%)
149088 bytes of code in asm (10.5439%)

41547 total symbols
6760 symbols documented (16.2707%)
34787 symbols undocumented (83.7293%)

14386873 total bytes of data
1264037 bytes of data in src (8.7860%)
13122836 bytes of data in data (91.2140%)
9868500 bytes of data in 29337 incbins (68.5938%)
```

PR #374, opened 2026-08-03
match remaining MIDI songs and align mid2agb with MID2AGB 1.06a

jiangzhengwenjz:
AI-Assisted-by: ChatGPT (GPT-5.6 Thinking)

# Match the remaining MIDI songs and align `mid2agb` with MID2AGB 1.06a

## Summary

This PR removes the final hand-written song assembly files and makes all 18 MIDI files that were previously under `sound/songs/midi/non_matching/` reproduce the original linked M4A data exactly.

It also brings the open-source `mid2agb` implementation into line with the official **MID2AGB 1.06a** converter for normal supported inputs, while adding explicit reconstruction metadata for information that cannot be represented by an ordinary dumped MIDI alone.

After this change:

- the former 18 nonmatching MIDI files generate the exact original relocated song bytes;
- their hand-written `sound/songs/*.s` replacements are removed;
- `linker.ld` uses the MIDI-generated objects;
- all 424 KATAM MIDI files match official MID2AGB 1.06a output under the official option set; and
- malformed inputs are handled without reproducing the official binary's unsafe memory behavior by default.

## Why both converter work and MIDI changes are necessary

The remaining songs were not primarily wrong at the musical level. The missing information was conversion structure.

M4A bytecode does not map uniquely back to MIDI. Equivalent playback can be represented by different:

- `Wxx` decompositions;
- event orders at the same timestamp;
- running-status decisions;
- measure and pattern boundaries;
- `PATT` definitions and references; and
- per-track loop targets.

A ROM-to-MIDI dump therefore preserves notes and audible controller data, but may lose authoring details that affected the original converter. Reproducing the original bytes requires two complementary changes:

1. make the converter follow official MID2AGB behavior for general MIDI processing; and
2. record irrecoverable song-specific control-flow or encoding information explicitly in the MIDI/configuration source.

The converter core does not contain per-song checks.

## Official MID2AGB compatibility

The official `mid2agb.exe` from AGBSDK 4.0 identifies itself as **MID2AGB 1.06a**. It was used as a behavior oracle through a controlled 32-bit PE execution environment.

The general converter changes reproduce observed official behavior in the following areas:

- numeric option parsing and output-label derivation;
- MIDI header and format interpretation;
- event classification and running status;
- note-on/note-off pairing, retrigger handling, and gate duration calculation;
- controller aliases and extended M4A commands;
- stable same-tick event ordering;
- coalescing adjacent same-tick commands when the final value supersedes the earlier one;
- time conversion and wait splitting;
- tie creation and compression scoring;
- pattern discovery and reuse; and
- loop-label and control-flow emission.

These are general rules derived from the official executable and cross-checked against real MIDI corpora. They are not modifications made only to force the 18 KATAM songs to match.

## Safety and robustness rationale

The official executable is used as the reference for valid conversion behavior, not as a requirement to copy unsafe implementation bugs.

The parser now bounds all reads and skips against both the physical file and the declared MIDI track, validates MIDI data bytes before table indexing, checks arithmetic and synthetic-event growth, and removes incomplete output after conversion errors.

This prevents malformed files from causing:

- out-of-bounds reads or writes;
- invalid controller-table indexing;
- unbounded timing expansion;
- signed-overflow-dependent behavior;
- use-after-free or double-free paths; or
- leaked partially generated output files.

### `-U`: opt-in reproduction of one known official bug

SA2's `mus_final_ending.mid` contains an overlong VLQ sequence that makes MID2AGB 1.06a wrap a timestamp and index outside its duration table. The official program consequently emits two very large wait values.

The safe default rejects the resulting non-chronological event stream. Passing `-U` reproduces that one known output deliberately and without performing an actual out-of-bounds access or signed overflow.

This option is isolated, disabled by default, and is not used by KATAM.

## Reconstruction-only extensions

The following options and source markers are intentionally separate from official MID2AGB compatibility. They describe information present in the original KATAM M4A tracks that was lost when the tracks were dumped to ordinary MIDI.

### `-F`: unresolved deferred-loop fallback

If a deferred loop does not encounter its preferred controller, tempo, or instrument anchor, place it before the first note. This recovers tracks whose original loop begins at the first musical event.

### `-I`: enable encoding-only MIDI markers

Interpret non-audible text/controller markers used only to reconstruct command ordering and wait/pattern decisions. With `-I` disabled, these markers do not change standard conversion behavior.

### `-Jtrack:offset`: exact per-track `GOTO` destination

Use `track base + byte offset` as the selected track's loop target. A sequence-level MIDI loop marker cannot express all original per-track M4A entry points.

### `-Q`: protect control-flow boundaries from pattern compression

Do not use a measure beginning immediately after a loop or label as a pattern definition/match start. Absorbing such a boundary into `PATT` can remove or move a required branch target.

### `-Strack`: suppress loops on selected tracks

Some MIDI files contain a sequence-level loop marker even though individual original M4A tracks terminate with `FINE`. This option removes the loop from those tracks before compression.

### `-Ttrack:ticks`: delay a selected track's loop end

Some tracks reach the original `GOTO` slightly later than the global sequence marker. The delay is applied before event ordering and wait splitting.

### `]+`: order loop end after same-tick controllers

This marker has the same loop meaning as `]`, but sorts after controller events at the same timestamp.

### `!`: reset emitted command state

This marker emits no M4A command. It forces the next operation to include its opcode rather than using the otherwise valid command-state abbreviation.

## MIDI and build integration rationale

The corrected MIDI files include explicit measure boundaries and, where necessary, non-audible timing hints. These do not change note timing or playback; they restore the converter's original wait and pattern decision boundaries.

Song-specific reconstruction settings are stored in `sound/songs/midi/midi.cfg` rather than hidden in generated assembly. This keeps exceptional control flow visible and reproducible.

The 18 MIDI files are moved into the normal MIDI directory, their generated hand-written assembly substitutes are removed, and `linker.ld` is changed to link the normal MIDI object paths. This makes MIDI the editable source of truth for the entire sound library.

## Validation

The final source tree was checked as follows:

- **KATAM official-option corpus:** 424/424 MIDI files generate output identical to MID2AGB 1.06a.
- **Former nonmatching songs:** 18/18 generate identical fully linked `.rodata`, including resolved `GOTO` and `PATT` destinations.
- **External real-world corpus:** 1,590/1,591 FE8U, pokeemerald, and SA2 MIDI files match the official converter in safe default mode. The only rejection is SA2 `mus_final_ending`, the known official overflow/OOB case.
- **Unsafe compatibility mode:** 1,591/1,591 external MIDI files match the official converter with `-U`.
- **Sanitizers:** all 1,591 external real MIDI files pass ASAN, UBSAN, and LeakSanitizer; the known SA2 file is safely rejected in default mode.
- **Malformed-input testing:** 46 directed malformed cases and 600 random mutations (646 total) pass sanitizer testing in both default and `-U` modes.
- **Build warnings:** `mid2agb` builds successfully with the project's `-Werror` configuration.

## Scope of the result

The official compatibility claim applies to MID2AGB 1.06a behavior exercised by KATAM and the tested real/synthetic corpora. Reconstruction-only options remain necessary because no converter can infer authoring metadata that is absent from a dumped MIDI.


ultra-azu:
I tried looking really hard at it and don't think I have any notes with this one, good job. At the end, for the scope of what we want it makes sense to hardcode some GOTOs and whatnot.

ultra-azu (approved):

jiangzhengwenjz:
```
1424856 total bytes of code
```

jiangzhengwenjz:
```
41770 total symbols
    7015 symbols documented (16.7944%)
    34755 symbols undocumented (83.2056%)

14512864 total bytes of data

```

PR #375, opened 2026-08-03
Name 18 object-engine symbols, with a manifest

jackmgaller:
The shared object/sprite engine API that nearly every object file calls into — `FindClosestKirby`, `ObjectSetBounds`, `ClearObjectBase`, `RequestScreenShake`, `GetCollisionTile` and friends.

Doing these first keeps the per-object naming PRs that follow free of cross-cutting renames; otherwise every boss PR wants to rename `sub_0803E308` in its own file and conflicts with the next one.

137 files, 3,843 lines changed of which 3,843 are removals — a pure rename is exactly symmetric. All six ROMs match from a clean build.

**Three symbols were audited and deliberately left as `sub_`/`gUnk_`**, including `sub_0803E2B0`, which is the most-referenced of the set at 78 files. Reasons per symbol are at the end of the manifest; the short version is that `sub_0803E2B0`'s fields have no readers in decompiled C, so I could not tell a damage hitbox from a solid body and would rather not guess.

**Ordering:** this touches no file that #373 adds, so the two are independent, but whichever lands second will need a trivial rebase. Happy to sequence them however you prefer.

AI-assisted (Claude).


jackmgaller:
Naming manifest attached here rather than committed to the tree.

**Downloadable file:** https://gist.github.com/jackmgaller/499857024f074908eaecca4ff097ac61

Inline copy below for convenience.

<details>
<summary><b>Naming manifest — object engine core (18 renames)</b></summary>

# Naming manifest — object engine core

18 renames. These are the shared object/sprite engine entry points that almost
every object file calls into. Doing them as one PR keeps the per-object PRs that
follow free of cross-cutting renames — otherwise every boss PR would want to
rename `sub_0803E308` in its own file and conflict with the next one.

137 files touched, 3,843 lines changed, of which 3,843 are removals — a pure
rename has an exactly symmetric diff. All six ROMs match from a clean build.

Line numbers refer to upstream at `c859c3b`.

**Three symbols were audited and deliberately left alone.** They are listed at
the end with what stopped me. One of them, `sub_0803E2B0`, is the most-referenced
symbol in the set (78 files) — I would rather ship 18 names than guess at a 19th.

---

## Kirby / player globals

| old | new |
|---|---|
| `gUnk_0203AD3C` | `gLocalPlayerId` |

`code_080332BC.c:38` sets it to `(*(vu32*)REG_ADDR_SIOCNT << 0x1a) >> 0x1e` —
bits 4–5 of `SIOCNT`, which on GBA hardware is the **multiplayer ID of this
console** in a link session. Every other writer agrees: `multi_08019F28.c:314`
and `:683` set it to `0` on the single-system paths, and `:362` takes it from
`gMultiBootStruct.unk00`. Readers use it as `gKirbys[gLocalPlayerId]` and, in the
`PlaySfx` macro (`data.h:9`), to decide whether a sound belongs to this console's
view.

*Note:* the earlier draft of this called it `gCurrentPlayerId`. "Current" reads
like a loop index; this is specifically *this console's* id, hence `Local`. Happy
to use either.

| old | new |
|---|---|
| `gUnk_0203AD44` | `gNumKirbys` |
| `gUnk_0203AD30` | `gNumPlayers` |

Both bound loops over `gKirbys[]`, and `FindClosestKirby` uses them together:

```c
for (i = 0; i < gNumKirbys && (i < gNumPlayers || r7 == INT_MAX); ++i)
```

so `gNumPlayers <= gNumKirbys`. `gNumKirbys` is set to `4` wherever a session
starts (`multi_08019F28.c:315`, `:684`, `subgame_menu.c:891`) and `0` at init
(`code_080332BC.c:35`). `gNumPlayers` is set to `1` on single-system paths
(`multi_08019F28.c:316`, `:685`) and to `gMultiBootStruct.unk01` on linked ones —
i.e. human consoles, against four Kirbys.

| old | new |
|---|---|
| `sub_0803D368` | `FindClosestKirby` |

`kirby.c:5xxx`. Loops over Kirbys in the same room, computes
`abs(dx) + abs(dy)`, keeps the smallest, returns `struct Kirby *`. The name
restates the body exactly.

---

## ObjectBase lifecycle

| old | new |
|---|---|
| `sub_0803E380` | `ClearObjectBase` |
| `sub_0803E3B0` | `ClearObject4` |

`ClearObjectBase` is `CpuFill16(0, r4, sizeof(struct ObjectBase))` plus explicit
resets of `unk58`, `unk6C`, `parent`, `sprite.tilesVram` and
`sprite.unk20[0].unk0`. `ClearObject4` is `CpuFill16(0, r1, sizeof(struct Object4))`
and nothing else. The two differ only in which struct they clear, which is what
the names say.

| old | new |
|---|---|
| `sub_0803DCCC` | `ObjectBaseDestroy` |

Takes a `struct Task *` and is passed as the `TaskCreate` destructor throughout
(e.g. `code_0806F780.c:10415`, `:10879`). It frees the sprite's VRAM
(`VramFree(...tilesVram)` when the allocation is above `0x6014000` and the
`0x4000` flag is clear), branching on `unk0 == 3` to read the fields at their
`Object4` offsets instead. Consistent with the existing `ObjectDestroy`.

| old | new |
|---|---|
| `sub_0803E308` | `ObjectSetBounds` |

Stores its four arguments raw into `unk3C`–`unk3F`. Those fields are read by the
**terrain collision** code: `code_080023A4.c:721`, `:748`, `:774`, `:800` compute
`arg0->x + arg0->unk3E * 0x100` to derive tile coordinates, `:2032`–`:2034`
derive tile columns from `unk3C`/`unk3E`, and `ability_objects.c:166` clamps
`x + unk3E` against `gCurLevelInfo[...].unk_S32Vec2_74.x`, a level edge. So this
is the box tested against the map, not against other objects.

---

## Object update and drawing

| old | new |
|---|---|
| `sub_0806F780` | `ObjectPreUpdate` |

`code_0806F780.c:724`. The guard every object's main runs first. Returns `TRUE`
when the caller should not continue: it destroys the task on the `0x1000` flag,
sets `0x1000` when the object has left its room or the room is gone, and when the
player is paused it inserts the object into the draw list and returns `TRUE`
without updating. Call sites read `if (!ObjectPreUpdate(objBase)) { ...update... }`.
**The return value means "already handled, stop"** — worth knowing before reading
the name.

| old | new |
|---|---|
| `sub_0806FAC8` | `Object4PostUpdate` |
| `sub_0803DBC8` | `Object4DisplaySprite` |
| `sub_0815604C` | `DisplaySprite` |

`Object4PostUpdate` (`code_0806F780.c:776`) is the tail of an `Object4`'s frame:
it runs `Macro_080FC150`, adds the screen-shake offset `gUnk_0203AD18[0..1]` to
the sprite position, and hands off to `Macro_0803DBC8`.

`Object4DisplaySprite` (`kirby.c:5633`) converts world to screen —
`sprite->x = (r8->x >> 8) - (gCurLevelInfo[gLocalPlayerId].viewportPosition.x >> 8) + r8->objBase54`
— adds the same shake offset, and displays. It is the `Object4` counterpart of
the `ObjectBase` version.

`DisplaySprite` (`sprite_2.c`) takes a bare `struct Sprite *` and builds the
`OamData` / `SpriteAttributes` entries. It is the lowest level of the three.

| old | new |
|---|---|
| `sub_080708DC` | `ObjectBaseInitSprite` |
| `sub_080709F8` | `Object4InitSprite` |

`code_0806F780.c:1196` and `:1243` are the same function twice, differing only in
whether the first parameter is `struct ObjectBase *` or `struct Object4 *`. Both
take `(obj, sprite, vramOrTileCount, animId, variant, numTiles)`, allocate VRAM
when the third argument is a count rather than an address, and set up the sprite.
Matches the existing `ObjectInitSprite`.

| old | new |
|---|---|
| `sub_0808AE30` | `CreateEffectObject` |

`code_0806F780.c:10878`. Creates a child `Object4` task parented to the caller,
at the caller's position and room, with the sprite chosen by the caller's
arguments. Deliberately generic — it is a spawner, and the *caller* decides what
the effect is.

---

## Collision lookup

| old | new |
|---|---|
| `sub_080023E4` | `GetCollisionTile` |
| `gUnk_082D88B8` | `gCollisionAttributes` |

`code_080023A4.c:1272`:

```c
u8 sub_080023E4(u8 playerId, u16 x, u16 y)
{
    struct LevelInfo *var0 = gCurLevelInfo + playerId;
    return gUnk_02024ED0[var0->unk65E][var0->roomWidth / 16 * y + x];
}
```

A tile fetch by room-relative tile coordinate. Call sites then feed the result
straight into the attribute table —
`gCollisionAttributes[GetCollisionTile(...)] & 0x200` — which is what makes
`gUnk_082D88B8` an attributes table indexed by tile value.

*Not renamed here:* `sub_08002434` and `sub_08002470` are near-identical
neighbours (one takes a `LevelInfo *`, one a different index). They are used by
too few files to be in this PR, but if you would rather the family were named
together I am happy to add them.

| old | new |
|---|---|
| `sub_0806FE64` | `RequestScreenShake` |

The one worth reading the chain for. `code_0806F780.c:874` takes
`(u16 kindAndFlags, struct ObjectBase *)`, finds the per-room slot in the
`gUnk_03000518` task whose `roomId` matches the object, and records a magnitude
(`unk6`) and flags. Nothing in that function shakes anything.

`sub_08070404` (`:1056`) runs each frame over the four slots and calls
`sub_08070164` on any with the `0x100` flag. `sub_08070164` (`:1xxx`) indexes
`gUnk_08350BB0[unk4]` / `gUnk_08350BCC[unk4]` by a per-frame counter and writes
the pair into `gUnk_0203AD18[0]` and `[1]`.

Those two words are the payload:

```c
gCurLevelInfo[playerId].altViewport_34.x -= gUnk_0203AD18[0] * 0x100;   // code_080023A4.c:1910
```

and `Object4PostUpdate` / `Object4DisplaySprite` add them to every sprite
position. So the request drives a per-room camera offset from a table of
oscillating values — a screen shake, per player view. The `& 0xFF` part of the
argument selects the table, and the callers are all heavy impacts.

---

## Audited and deliberately left unnamed

| symbol | files | why not |
|---|---|---|
| `sub_0803E2B0` | 78 | Sets a second box next to `ObjectSetBounds` — `unk38`/`unk39` take the given corner and `unk3A`/`unk3B` take **half** the extents, which is the storage form you would choose for an overlap test. But `unk38`–`unk3B` have essentially no readers in decompiled C; the only one I found is `code_0806F780.c:7113`, where a rider's `yspeed` is adjusted by the platform's `unk3B`, which reads as a solid body rather than a damage hitbox. "Hitbox" and "body" are both live readings and I could not separate them, so it keeps its address. **If you know which it is, I will add it.** |
| `sub_08089864` | 16 | Spawns an `Object4` with animation `0x293` hardcoded, at `±a2` horizontally with an outward x velocity — consistent with an impact effect. But `0x293` is also spawned from Kirby code in non-impact contexts (`code_0806F780.c:3948`, `:4616`, `:4673`, `:2371`, `:5682`), so a name like `CreateImpactStars` would be asserting what the artwork looks like rather than what the code says. |
| `sub_08155128` | 23 | Still assembly — there is no body to read. Used as `if (!sub_08155128(sprite))` inside the sprite macros, so it plausibly steps an animation and reports completion, but that is inference from call sites alone. |


</details>

src/multi_08019F28.c
```diff
@@ -680,9 +680,9 @@ static void sub_0801A6BC(struct Multi_08019F28 *r5)
 void sub_0801A744(s32 r0)
 {
     gUnk_0203AD10 = 4;
-    gUnk_0203AD3C = 0;
-    gUnk_0203AD44 = 4;
-    gUnk_0203AD30 = 1;
+    gLocalPlayerId = 0;
+    gNumKirbys = 4;
+    gNumPlayers = 1;
```
jiangzhengwenjz: If `gNumPlayers` is the number of human players, then why is it not `gNumHumanPlayers`?

src/kirby.c
```diff
@@ -5328,14 +5328,14 @@ const struct AnimInfo gUnk_08350B30[] = {
...
-struct Kirby *sub_0803D368(struct ObjectBase *r6)
+struct Kirby *FindClosestKirby(struct ObjectBase *r6)
 {
     u8 i = 0; // redundant initialization
     u32 sp00 = 0, sl = 0;
     bool32 sp04 = TRUE;
     s32 r7 = INT_MAX;
 
-    for (i = 0; i < gUnk_0203AD44 && (i < gUnk_0203AD30 || r7 == INT_MAX); ++i)
+    for (i = 0; i < gNumKirbys && (i < gNumPlayers || r7 == INT_MAX); ++i)
```
jiangzhengwenjz: It can't always be closest. As you said `gNumPlayers` could be just 1 so the loop may not even check every kirby.

jiangzhengwenjz (changes requested):

jiangzhengwenjz:
Please resolve merge conflicts as #373 has been merged

src/kirby.c
jackmgaller: You're right, and this was the better catch of the two — `Closest` is wrong. Renamed to **`FindTargetKirby`**.

Reading the condition properly:

```c
for (i = 0; i < gNumKirbys && (i < gNumHumanPlayers || r7 == INT_MAX); ++i)
```

past index `gNumHumanPlayers` it continues **only while `r7 == INT_MAX`** — only until something is found. So it is closest-of only among the human-controlled Kirbys; beyond those it takes the *first* one in the room and stops. At `gNumHumanPlayers == 1` that means index 0 is measured and the rest are first-match, exactly as you said.

What does hold regardless: the result is always in the same room as the object, `unkC & 0x8000` candidates are held back in `sl` and used only if nothing else matched, and every caller stores it as the object's target (`obj->kirby3 = ...`). So `FindTargetKirby` says what it is for without claiming an ordering it does not guarantee.

If you would rather it stayed `sub_0803D368` until someone works out the intent behind the early exit, I am fine with that too.

src/multi_08019F28.c
jackmgaller: Fair — renamed to **`gNumHumanPlayers`**. If the justification is "human players" the name should carry it.

The distinction is already explicit nearby: `code_080332BC.c:43` sets it two lines above a read of `gAIKirbyState`, and the Kirbys from `gNumHumanPlayers` up to `gNumKirbys` are the AI ones — which is also why the two bounds appear together in `FindTargetKirby`.

jackmgaller:
Rebased onto `ec691f0` — conflicts resolved by rebuilding the branch on the new master and re-running the sweep, so the three files #373 added (`sword_knight.c`, `vert_sliding_door.c`, `eight_dir_cannon.c`) are covered too. Force-pushed `1862834`.

Both names corrected:

- `FindClosestKirby` → **`FindTargetKirby`** — you were right that it can't always be closest; past `gNumHumanPlayers` the loop stops at the first match.
- `gNumPlayers` → **`gNumHumanPlayers`**.

137 files, symmetric 3,831/3,831 diff, six ROMs verified from a clean build. [Manifest](https://gist.github.com/jackmgaller/499857024f074908eaecca4ff097ac61) updated with both corrections.

jiangzhengwenjz (approved):

jiangzhengwenjz:
```
1413968 total bytes of code
1264880 bytes of code in src (89.4561%)
149088 bytes of code in asm (10.5439%)

41547 total symbols
6778 symbols documented (16.3141%)
34769 symbols undocumented (83.6859%)

14386873 total bytes of data
1264037 bytes of data in src (8.7860%)
13122836 bytes of data in data (91.2140%)
9868500 bytes of data in 29337 incbins (68.5938%)
```

PR #376, opened 2026-08-04
Decompile hot_head, leap and cookin (44 functions)

sh1ftmaker:
Decompiles 3 enemy object files (44 functions) to C — all matching, nothing ships as NONMATCHING.

| File | Functions | Status |
|---|---|---|
| src/hot_head.c | 14 | all matching |
| src/leap.c | 12 | all matching |
| src/cookin.c | 18 | all matching |

No baserom on this machine, so each object is verified directly instead of `make compare`: compiled `.text` bytes and relocations are identical to the assembled original asm, leaving the linked ROM unchanged. CI's `make compare` should confirm.

AI-assisted (Claude).


jiangzhengwenjz:
Hello, I am glad to have more contributors. However, an excessively large PR makes the review process difficult, so please split your PR into smaller ones.
Apart from this, i hope you can guide Claude to match them better. Currently we are near 90% and there's only one function in `asm/nonmatching`.
In addition, I'll decompile `code_0806F734.s` later when I have time and Mateo has claimed `prank.s` so please leave them to us.

sh1ftmaker:
Sorry about the oversized PR — understood. Split into per-file PRs, keeping only files where every function matches:

- #379: hot_head, leap and cookin (44 functions, all matching)
- #377: code_08020220 (15 functions, all matching)
- #378: code_0802F8D8 (40 functions, all matching)

prank and code_0806F734 are dropped everywhere — I hadn't checked claims carefully enough before starting, sorry. Leaving both to you and Mateo; if the finished drafts are useful as reference just say so, otherwise ignore them.

The rest of what was in this PR leaned too hard on NONMATCHING wrappers. I'll bring those files up to real matches before submitting anything else, and keep future PRs small.

PR #378, opened 2026-08-04
Decompile code_0802F8D8 — sprite effect tasks (40 functions)

sh1ftmaker:
Decompiles `code_0802F8D8.c`, 40 functions / 1325 asm lines: the sprite effects belonging to `code_0802E57C` — two tasks that carry a sprite along a fixed velocity for a set number of frames, and the per-component state machines dispatched through `gUnk_082EB7D0`.

**Scope.** One file, plus the shared header it cannot compile without. `include/code_0802E57C.h` gains offset comments, a name for `unk48`, and the correct type for the `unk28` callback; `include/code_0802F8D8.h` is new and two declarations leave `functions.h`, which touches `code_08138D64.c`.

**Evidence.**

- The `unk28` signature. The dispatcher in `sub_0802EF90` loads r0 with the parent and r1 with the component before calling through the member (`ldr r2, [r4, #0x28]` / `adds r0, r6, #0` / `adds r1, r4, #0` / `bl _call_via_r2`), and this file assigns 21 two-argument functions to it across 22 sites. The old one-argument type could not express them, which is why the file previously carried a private copy of the struct and reached the same memory through casts. Both casts are gone.
- `unk48` is named because this file uses it ten times. The rest of the old `filler48` stays filler: there is assembly evidence for a `u16` at 0x4A — nineteen `ldrh`/`strh` accesses across `code_0802E57C.s`, `code_0802B4A8.s` and `cutscene_trigger.s`, two of them written as `adds #2` off 0x48 — but no C consumes it and this PR decompiles none of those files, so naming it belongs to whoever does.
- Linkage: 18 static, 22 global, counted from the built object. The split comes from the linker map plus a scan of `baserom.gba` for word pointers and Thumb BLs to each address. Five of the globals are reached only through `gUnk_082EB7D0`, which is still `.incbin`, so nothing in the tree names them and a repository grep alone would have made them static wrongly. `sub_0802FC08` is additionally named by a literal pool word in `asm/code_0802E57C.s`.

**Unusual source forms, and why.** Six, each under `#ifndef NONMATCHING` with a plain readable branch and a comment naming what it resolves: an r8 clobber in `sub_0802F8D8` that changes the allocation, four tied-operand barriers that fix the order two constants are materialised in, and an r0 pin in `sub_080300CC` where the clamp result lives in r0. `sub_0802F8D8` also passes `unk1B` through a statement expression because the original materialises -1 as a signed word; `0xFF`, `-1`, `(s32)-1` and an `s32` local each change the output.

**Left unresolved.** Every symbol in the file keeps its `sub_`/`gUnk_` name. The component struct's fields are still `unkNN`; nothing here justified renaming them.

**Validation.** Clean rebuild, `make -j8 compare` gives six `OK` lines: `unk_8D94B9C`, `unk_8E1FE28`, `unk_8E8490C`, `payload`, `speed_eaters`, `katam`. The whole-project `NONMATCHING` build links and produces a different ROM.

**Conflicts.** Clean against master, against #367, #374 and #380, and against #377 and #379.


sh1ftmaker:
I have applied the points from #377 to this PR rather than wait for the same review twice. Pushed as one commit; `make compare` still gives six `OK` lines.

**The private copy of `struct Unk_0802E57C_C` is gone.** This file used to define its own version with `filler48` named, and reach the real memory through two casts. The fields are named in `include/code_0802E57C.h` now — `unk48`, `unk4A`, `filler4C` — the callback member carries its real two-argument type through a typedef instead of a one-argument lie, and both structs in that header gained offset comments. No casts left.

**Sixteen functions are static.** I checked which ones can be by scanning the ROM for word-aligned pointers and Thumb `BL`s to each address, not by grepping the repo — six of them are dispatched from the table at `gUnk_082EB7D0`, which is `.incbin`'d, so nothing in the tree names them and grep alone would have made them static wrongly.

**`include/code_0802F8D8.h`** declares the two functions `code_08138D64.c` calls, and they are out of `functions.h`. The rest are reached from `asm/code_0802E57C.s` or through that table, so per the same reasoning as #377 they have no declaration.

**Unused parameters are `UNUSED`** — twenty of them — and the callback arguments are `arg0`/`arg1` rather than `a`/`b`.

**Every matching trick says what it is for now.** The `asm("":::"r8")` barrier, the `u8 variant` pins, the pointer aliases in the two task functions, the `r0` pin in `sub_080300CC`, and the one statement expression. On that last one: it is there because `unk1B` has to be materialised as a signed word. I tried `0xFF`, `-1`, `(s32)-1` and a typed local, and recorded in a comment what each does, so the next person does not have to repeat it.

sh1ftmaker:
Rebased onto master and force-pushed as a single commit, and reworked ahead of review rather than waiting for the same comments as #377.

**Corrections to what I wrote here last time.** I said sixteen functions were static; it is eighteen, counted from the built object — the sixteen was how many I had converted, not how many there are. I said six of the global ones are reached only through `gUnk_082EB7D0`; it is five, because `sub_0802FC08` is also named by a literal pool word at `0x0802EAF4` in `asm/code_0802E57C.s`.

### The shared header

`include/code_0802E57C.h` gains offset comments, a name for `unk48`, and the real type for the `unk28` callback. That last one is not cosmetic: the dispatcher in `sub_0802EF90` loads r0 with the parent and r1 with the component before calling through it —

```
	ldr r2, [r4, #0x28]
	adds r0, r6, #0      @ parent
	adds r1, r4, #0      @ component
	bl _call_via_r2
```

— and this file assigns 21 two-argument functions to that member across 22 sites. The old one-argument type could not express them, which is why the file previously defined a private copy of the struct and reached the same memory through casts. Both casts are gone.

The rest of the old `filler48` stays filler. There is evidence in assembly for a `u16` at 0x4A — nineteen `ldrh`/`strh` accesses across `code_0802E57C.s`, `code_0802B4A8.s` and `cutscene_trigger.s`, including two written as `adds #2` off 0x48, which is why a search for the offset alone misses them — but no C consumes it and this PR does not decompile any of those files, so naming it belongs to whoever does. `unk48` is named because this file uses it ten times.

### Linkage

18 of the 40 functions are static, 22 global. The split comes from the linker map plus a scan of `baserom.gba` for word pointers and Thumb BLs to each address. Five of the globals are reached only through `gUnk_082EB7D0`, which is still `.incbin`, so nothing in the tree names them and a repository grep alone would have made them static wrongly.

`include/code_0802F8D8.h` declares the two functions `code_08138D64.c` calls; both are out of `functions.h`.

### What I removed on a second pass

Thirty-odd lines that looked like matching work and were not. Ten function-pointer locals of the form `f = X; unk28 = f; f(arg0, arg1);`, three `u8 variant = 0;` locals at sites with no barrier, and a `u16 unk3A` struct member with no access anywhere. Each is byte-identical when deleted; `sizeof(struct Unk_0802F8D8)` is still 0x3C, which the `movs r1, #0x3c` before `TaskCreate` confirms.

I also rewrote four comments that described mechanisms I had guessed at. The `asm("":"+r"(variant))` barriers do not keep the variant in a register — it is in one either way — they fix the order two constants are materialised in, which the diff shows as `movs r2,#0 / movs r0,#0` swapping. The pointer alias in `sub_0802F9AC` does not split a value across two registers; it forces the extra `adds r4, r0, #0` the original has.

### Matching constructs that remain

Six, each under `#ifndef NONMATCHING` with a plain readable branch and a comment saying what it resolves: an r8 clobber in `sub_0802F8D8`, four tied-operand barriers, and an r0 pin in `sub_080300CC` where the clamp result lives in r0. `sub_0802F8D8` also passes `unk1B` through a statement expression because the original materialises -1 as a signed word; `0xFF`, `-1`, `(s32)-1` and an `s32` local were each tried and each changes the output.

### Validation

Clean rebuild, `make -j8 compare` gives six `OK` lines: `unk_8D94B9C`, `unk_8E1FE28`, `unk_8E8490C`, `payload`, `speed_eaters`, `katam`. The whole-project `NONMATCHING` build links and produces a different ROM.

jiangzhengwenjz:
If you still want to get this and #379 merged, please feed AGENTS.md to Claude, update the PRs, co-author Claude and make sure they are ahead of `master`.

PR #379, opened 2026-08-04
Decompile the hot_head, leap and cookin enemies (44 functions)

sh1ftmaker:
Decompiles three enemies, 44 functions: `hot_head.c` (14 functions / 2053 asm lines), `leap.c` (12 / 1879) and `cookin.c` (18 / 1885). Every function matches.

**Scope.** Three files that are one kind of work and share no state; each gets its own header. `sub_080C3694`'s declaration leaves `functions.h` — it was the only one there, the other ten globals were declared nowhere — which touches `shadow_kirby.c`. `prank.h` gains `sub_080A158C` and `sub_080A245C`, which `cookin.c` calls.

**Evidence.**

- Linkage: 33 static, 11 global, counted from the built objects. From the linker map plus a scan of `baserom.gba` for word pointers and Thumb BLs to each address. `sub_080C3694` is BL'd from `shadow_kirby.c`; `sub_080B0758` and `sub_080B6368` from `asm/code_08032E98.s`; the remaining globals are named by ROM object tables. Taking extents from the map rather than from `sub_` symbol names matters for `leap.o`, which begins at `CreateLeap`: `sub_080B0860` is called only from there, so a name-derived extent puts its one caller outside the file.
- `SetPointerSomething` guarded by `!(flags & 0x200)` replaces the private near-copy `hot_head.c` carried and the same body `leap.c` had written out by hand. Guarded and unguarded both reduce to `!(flags & 0x1200)`, and this is the idiom already in `boxin.c` and `box_boxer.c`.
- `dirIndex`: `gUnk_0835553A` is six `s16` pairs — (756,112) (759,-72) (759,72) (756,-112) (768,0) (0,814) — five of constant magnitude at different angles, so it is a direction table, and the value is clamped to 5 before indexing it.
- `cookin.c` uses `OBJ_COOKIN`, `OBJ_EMPTY_31` and `OBJ_COOKIN_PAN` from `constants/object_types.h` in place of `0x26`, `0x31` and `0x9E`.
- `sub_080A245C` is still in `asm/prank.s`; `sub_080A158C` is in `src/prank.c`.

**Unusual source forms, and why.** Three. `cookin.c` pins r0/r1/r2 around two `ObjectSetFunc` calls, under `#ifndef NONMATCHING` with a plain readable branch, because the original sets the arguments up r2, r0, r1 where the plain call does r1 first — the construct `dark_meta_knight.c` already uses. `leap.c` keeps a `u16 zero = 0` local because the original ORs a register known to be zero into flags. `hot_head.c` keeps `gUnk_0835553A[idx = dirIndex * 2]`, because hoisting the assignment reorders the two loads.

**Left unresolved.** Most locals and every function keep generic names. `gUnk_0835553A` is shown above to be a direction table but is not renamed here. `leap.c`'s spawned object is `objBase` rather than a role name: it never receives a sprite, only a collision box, so there was no evidence for one.

**Validation.** Clean rebuild, `make -j8 compare` gives six `OK` lines: `unk_8D94B9C`, `unk_8E1FE28`, `unk_8E8490C`, `payload`, `speed_eaters`, `katam`. The whole-project `NONMATCHING` build links and produces a different ROM.

**Conflicts.** Clean against master, against #367, #374 and #380, and against #377 and #378.


sh1ftmaker:
Same self-review as #378, applying what came out of #377. One commit, `make compare` still six `OK` lines.

**Ten functions are static**, chosen by scanning the ROM for pointers and Thumb `BL`s to each address rather than by grepping. That mattered here: `sub_080C3F2C` and `sub_080B0860` look internal — nothing in the tree names them — but both are called from code that is still in asm, so they stay global. Note the object files differ from the reference after this change even though the ROM does not: a `bl` to a global symbol leaves a relocation while a `bl` to a static one is resolved by the assembler. The linked ROM is identical, which is what `make compare` checks.

**`hot_head.c` no longer keeps its own copy of `SetPointerSomething`.** It is a twenty-line duplicate of the one in `object.h` differing only in how the flag test is spelled, so it lives in `object.h` now as `SetPointerSomethingAlt`, next to its sibling, with a comment saying the two are semantically identical — `!(f & 0x1200)` already implies `!(f & 0x200)` — but emit their branches in different orders, so they are not interchangeable.

**`leap.c` had the same body written out by hand**, nested four `if`s deep. It uses the macro now: seventeen lines shorter and byte-identical.

**Declarations moved to where they belong.** `sub_080C3694` is out of `functions.h` and into `hot_head.h`; the two prank functions `cookin.c` was declaring for itself are in `prank.h`, with a note that `sub_080A245C` is still in `asm/prank.s`.

**`cookin.c` uses `OBJ_COOKIN` and `OBJ_EMPTY_31`** instead of `0x26` and `0x31` — the constants were already in `constants/object_types.h` and I had missed them. Its register pins are behind `NONMATCHING` with a plain fallback, the way `dark_meta_knight.c` does it, and say what they are for. The whole-project `NONMATCHING` build compiles and produces a ROM that differs from the matching one, so those `#else` paths are real rather than dead text.

**`hot_head.c`'s Kirby argument is `kirby`**, which is what the rest of the repo calls it; I had called it `target`, which asserted something the code does not say.

sh1ftmaker:
Rebased onto master and force-pushed as a single commit, reworked ahead of review.

**Correction to what I wrote here last time.** I said twelve functions are static. It is thirty-three of forty-four, counted from the built objects; twelve was how many I had converted, not how many there are. I also said `sub_080C3F2C` and `sub_080B0860` had to stay global because assembly calls them. That was wrong, and the mistake is worth stating because it is not obvious: I derived each file's address range from its `sub_XXXXXXXX` symbols, but `hot_head.o` begins at `CreateHotHead`, 0x174 bytes earlier, so a caller inside the file looked external. Extents now come from `katam.map`. Both functions are static and the ROM is unchanged.

Of the five functions I originally listed as affected by that mistake, only `sub_080B0860` actually is — it is called from `CreateLeap` alone. The other four are called from ordinary `sub_`-named functions well inside the range and would never have been misclassified. The commit message says only what is true.

### `SetPointerSomethingAlt` is gone

`hot_head.c` had a twenty-line private near-copy of `SetPointerSomething`, and I had moved it into `object.h` as a variant. That was unnecessary. The idiom already in the repo — `if (!(objBase->flags & 0x200)) SetPointerSomething(objBase);`, used five times across `boxin.c` and `box_boxer.c`, and once in `cookin.c` in this PR — is byte-identical in both files. Guarded and unguarded both reduce to `!(flags & 0x1200)`, since `!(f & 0x1200)` already implies `!(f & 0x200)`.

So `include/object.h` is untouched by this PR now, and `leap.c`, which had the same body written out by hand four `if`s deep, calls the macro.

### Linkage

33 static, 11 global. From the linker map plus a scan of `baserom.gba` for word pointers and Thumb BLs to each address. `sub_080C3694` is BL'd from `shadow_kirby.c`; `sub_080B0758` and `sub_080B6368` from `asm/code_08032E98.s`; the rest of the globals are named by ROM object tables.

`sub_080C3694`'s declaration is out of `functions.h` — it was the only one there; the other ten globals were declared nowhere. `prank.h` gains `sub_080A158C` and `sub_080A245C`, which `cookin.c` calls; `sub_080A245C` is still in `asm/prank.s`.

### Style and naming

The three files disagreed with each other, which was mine to fix rather than yours to point out. `hot_head.c` and `leap.c` now use the same brace style as `cookin.c` and as the thirteen `.c` files added to `src/` this year, which are 105 same-line function definitions to zero. Same for `(struct Object *arg0, u8 arg1)` over `(template, a2)`, `struct Task *task` over `*t`, and the address form for the one `goto` label.

`cookin.c` uses `OBJ_COOKIN` and `OBJ_COOKIN_PAN` instead of `0x26`, `0x31` and `0x9E`, including in `gUnk_08351648[OBJ_COOKIN]`. In `leap.c` the spawned object is `objBase`, not `shadow` — it never gets a sprite, only a collision box, so "shadow" was an assertion about art nobody looked at. The one named from a sound constant is `kiss`, from `SE_LEAP_KISS_ATTACK`, rather than `heart`, which was a further guess on top of it. `hot_head.c`'s Kirby argument is `kirby`, matching the rest of the repo.

`dirIndex` I kept, and it is checked rather than assumed: `gUnk_0835553A` is six `s16` pairs — (756,112) (759,-72) (759,72) (756,-112) (768,0) (0,814) — five of constant magnitude at different angles, so it is a direction table and the value is clamped to 5 before indexing it.

### Matching constructs that remain

Three. `cookin.c` pins r0/r1/r2 around two `ObjectSetFunc` calls, under `#ifndef NONMATCHING` with a plain readable branch, because the original sets the arguments up r2, r0, r1 where the plain call does r1 first — the same construct `dark_meta_knight.c` uses. `leap.c` keeps a `u16 zero = 0` local because the original ORs a register known to be zero into flags. `hot_head.c` keeps `gUnk_0835553A[idx = dirIndex * 2]`, because hoisting the assignment reorders the two loads.

### Validation

Clean rebuild, `make -j8 compare` gives six `OK` lines: `unk_8D94B9C`, `unk_8E1FE28`, `unk_8E8490C`, `payload`, `speed_eaters`, `katam`. The whole-project `NONMATCHING` build links and produces a different ROM.

PR #371, opened 2026-08-02
Name 1792 symbols across the object, Kirby and menu code

jackmgaller:
Renames 1456 `sub_XXXXXXXX` functions, 329 `gUnk_XXXXXXXX` data symbols and 7 struct types, derived from function-pointer tables, state assignment graphs, SFX constants and call-site context. Purely textual — all six ROMs match from a clean build.

This work was AI-assisted.

Issue #64, opened 2021-03-04
TODO: detecting potential UBs


jiangzhengwenjz:
From my observation, a lot of `<<` operations are performed on (possibly) negative integers which is undefined behavior in C (apparently this also includes `s8` and `s16` because of integer promotion). 

This kind of UB should be eliminated as long as multiplication also matches the vanilla assembly code.

PR #380, opened 2026-08-05
Add AGENTS.md — instructions for AI agents

jackmgaller:
As discussed. A single file that AI assistants are pointed at, so the same corrections stop coming back to you.

**Everything in it is drawn from review comments you have already made on this repo** — I read through the inline comments across ~90 PRs and collected the recurring ones. Nothing is new policy, and I have deliberately not invented anything you have not asked for.

What it covers:

- **Pull requests** — small, one system at a time, wait for a merge before the next. Size guidance as a table, framed around how many *decisions* a reviewer has to check rather than lines changed. Six `OK` lines from `make compare`. No editor or tooling files.
- **Naming manifests** for renames — old name, new name, concrete evidence; attached, not committed. From your #371 comment.
- **Naming** — PascalCase / camelCase / `sPascalCase`, `unkNN` vs `fillerNN`, header guards. Plus the two failure modes you caught on me recently: a name that adds no knowledge, and a name that claims more than the code does.
- **Where code goes** — static functions out of headers, `functions.h`/`data.h` sorted by address, includes in the `.c`, newline at end of file, struct size comments. Also the `gUnk_08351648` case, since "no caller in this file" does not mean "static".
- **Types and constants** — use `sizeof`, the io_reg/palette/songs constants, the existing macros, `bool32` only when it really is a boolean, no unnecessary casts, `({ })` over `do {} while (0)`.
- **Formatting** — the spacing rules, switch indentation, tabs in `.s`.
- **Matching** — fake matches get a comment, the NONMATCHING hatch with its decomp.me link, and the reminder that `make compare` cannot see the `#else` body.

Documentation only, no code touched.

Two things I would especially like your view on: whether the size guidance is roughly right, and whether anything in it misstates what you actually want. Happy to cut or reword anything.

AI-assisted (Claude).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,230 @@
+# Instructions for AI agents
+
+If you are an AI coding assistant working on this repository, read this first and
+follow it. It exists so that reviewers spend their time on the decompilation
+itself and not on the same corrections over and over.
+
+Everything here comes from review comments already made on this repo. Nothing in
+it is new policy.
+
+---
+
+## The one rule that matters most
+
+**A wrong name or a wrong guess is worse than no name and no guess.**
+
+If you are not confident, leave the symbol as `sub_08XXXXXX` / `gUnk_08XXXXXX`,
+leave the type as it was, and say in the pull request what you were unsure about.
+Nobody minds an honest gap. Everybody minds a confident mistake, because it
+misleads every person who reads the file afterwards.
+
+---
+
+## Pull requests
+
+**Keep them small and submit them one at a time.** A pull request that a person
+cannot review in one sitting will not get reviewed. If your work is large, split
+it. Wait for one to be merged before opening the next, so you do not create merge
+conflicts for yourself.
+
+**One system per pull request.** Do not mix a rename of Kirby code with a rename
+of menu code. Do not mix decompilation with unrelated renaming.
+
+**Rough size guidance:**
+
+| Kind of change | Aim for |
+|---|---|
+| Decompiling | one file, or a few small related files |
+| Renaming | one system, about 20–40 symbols |
+| A fix or a match | as small as it can be |
+
+These are guidelines, not hard limits. A pure rename touching many files is fine
+if the number of *decisions* is small — what makes a pull request reviewable is
+how many judgements a person has to check, not how many lines changed.
+
+**Before you open a pull request:**
+
+- `make compare` passes and prints **six** `OK` lines: `katam.gba`, `payload.gba`,
+  `speed_eaters.gba`, `unk_8D94B9C.gba`, `unk_8E1FE28.gba`, `unk_8E8490C.gba`.
+  Check for all six. A partial pass looks very like a full one.
+- Do not commit editor or tooling files: `.clangd`, `compile_flags.txt`, personal
+  `.gitignore` entries, scratch scripts.
+- Say plainly in the description that the work was AI-assisted.
+
+**If you are renaming symbols, include a manifest.** For every rename: the old
+name, the new name, and the concrete evidence. "Concrete" means something a
+reviewer can check — a line number, a called function, a value written to a
+register. Not "it looks like a jump".
+
+Attach the manifest to the pull request. Do not commit it to the repository.
+
+**Do not rebase or force-push other people's branches.** If you must force-push
+your own, say so in a comment.
+
+---
+
+## Naming things
+
+| What | Style | Example |
+|---|---|---|
+| Functions | PascalCase | `CreateBonkers` |
+| Local variables | camelCase | `objBase`, `saveBuffer` |
+| Static variables | `s` + PascalCase | `sAreaMapRoomInfos` |
+| Header guards | `GUARD_<FILE>_H` | `GUARD_WADDLE_DEE_H` |
+| Unknown field, type known | `unk` + hex offset | `unk3C` |
+| Unknown field, type unknown | `filler` + hex offset | `filler6` |
+| Unknown local | `sp` + hex offset, or the register | `sp00`, `r4` |
+
+**A name has to add knowledge.** `gBonkersAnimInfo2`, `gBonkersAnimInfo3` tell a
+reader nothing that `gUnk_08356058` did not. If the only thing you can express is
+which order they appear in the file, leave them alone.
+
+**A name must not claim more than the code does.** Words like *Closest*, *First*,
+*All* and *Always* are claims about control flow. Check them against the loop's
+exit condition, not just its body. A function named `FindClosestKirby` that stops
+at the first match once it is past a certain index is a wrong name.
+
+**Match the words already used in the repository.** If the tree says `Eight`, do
+not introduce `8Dir`. If a struct field is documented one way, follow it.
+
+**Write real English.** The plural of *life* is *lives*. Keep capitalisation
+consistent in comments — pick `Dark Mind` or `DarkMind` and stay with it.
+
+**Be careful naming things after sound effects.** Check how many callers a sound
+constant has first. `SE_BOSS_GROUND_POUND_ATTACK` is played by eight different
+bosses on any heavy landing, so it identifies an impact, not an attack. A
+constant used by one or two callers is good evidence; a shared one is not.
+
+---
+
+## Where code goes
+
+**Static functions do not belong in headers.** Mark them `static` and forward
+declare them at the top of the `.c` file. Only symbols used by another file go in
+a header.
+
+**But do not mark something `static` just because nothing in the file calls it.**
+Some functions are reached only through a data table that is still raw binary —
+`gUnk_08351648` is the common one. Before making a function `static`, search the
+ROM for its address with the Thumb bit set (`address | 1`, 4-byte aligned). A hit
+outside its own literal pool means something references it.
+
+**If a declaration has no dedicated header yet, put it in `functions.h`** (for
+functions) or `data.h` (for data). Keep both sorted by address.
+
+**Includes belong in the `.c` file**, not in the header, unless the header itself
+needs the type.
+
+**Do not declare a symbol that is already declared somewhere else.**
+
+**End every file with a newline.**
+
+**Give each struct a size comment** so identical structs are easy to spot:
+
+```c
+struct VertSlidingDoor {
+    struct Object2 obj2;
+    u16 unkB4;
+    s8 unkB6;
+    s8 unkB7;
+}; /* size = 0xB8 */
+```
+
+---
+
+## Types and constants
+
+**Use the constants that already exist.** `DISPCNT_*`, `BLDALPHA_BLEND`,
+`RGB_WHITE`, `KEYS_MASK` and the rest are in `gba/io_reg.h`, `palette.h` and
+`include/constants/`. Sound effects are named in `include/constants/songs.h`.
+
+**Use `sizeof`** instead of writing the size as a number.
+
+**Use the existing macros** rather than open-coding what they do:
+`TaskGetStructPtr`, `PlaySfx`, the `SpriteInit*` family, `Macro_*`. If a macro
+almost fits but not quite, it is usually better to add a parameter to the macro
+and update its call sites than to work around it locally.
+
+**Use the most specific type you know.** If a pointer is really a
+`struct Object2 *`, say so rather than leaving it as `struct ObjectBase *`.
+
+**Use `bool32` with `TRUE` and `FALSE` when a value is genuinely a boolean** — but
+only then. If you are not sure a function returns a boolean, leaving it as `u32`
+is better than guessing. Write `if (x)`, not `if (x != FALSE)`.
+
+**Write `void` explicitly** for functions that take no arguments.
+
+**Do not add casts that are not needed.**
+
+**A multi-statement macro should use a statement expression `({ ... })`**, not
+`do { ... } while (0)`. It behaves the same and usually does not change codegen.
+
+---
+
+## Formatting
+
+- Space after `if`, `while`, `for`, `switch`, `do`.
+- Space between `)` and `{`.
+- Space before `?` and `:`.
+- Space after `//`.
+- `switch` cases are not indented relative to the `switch`.
+- No stray blank lines, and no more than one in a row.
+- Assembly files use tabs to indent.
+
+```c
+    switch (obj->subtype) {
+    case 0:
+        DoSomething(obj);
+        break;
+    default:
+        break;
+    }
+```
+
+---
+
+## Matching
+
+**A real match is always better than a fake one.** If you write something
+non-obvious purely to make the compiler produce the right instructions, add a
+comment saying what the natural form would have been.
+
+**If it will not match, use the sanctioned escape hatch** rather than grinding.
+Put a decomp.me link and the percentage above it, as the existing ones have:
+
+```c
+// sub_08153184: https://decomp.me/scratch/z1VmL (95.20%)
+#ifndef NONMATCHING
+NAKED bool32 sub_08153184(void) {
+    asm(".include \"asm/nonmatching/sub_08153184.inc\"");
+}
+#else
+bool32 sub_08153184(void) {
+    /* readable best-effort C */
+}
+#endif
+```
+
+Keep the `#else` body honest and readable. People work on these in decomp.me
+scratches, so a body that is close and clear is worth more than one that is
+slightly shorter and full of tricks.
+
+**`make compare` does not check the `#else` body.** That code is not built in the
+normal build, so six green ROMs say nothing about it. If you change a file
+containing one, build it separately with `NONMATCHING` defined and confirm the
+body still compiles and has not grown.
+
+**Measure one change at a time.** If an edit touches several places, test each
+place separately. A single global search-and-replace followed by one build gives
+you one answer to several questions, and it will be the wrong answer.
+
+**When you claim something does not match, show the difference** — the
+instructions that differ, not just "it failed". It is much easier to discuss.
+
+---
+
+## When you are unsure
+
+Ask in the pull request. Say what you checked, what you found, and what you could
+not resolve. A question with evidence attached is welcome. A guess presented as a
+fact is not.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -10,7 +10,7 @@ The purpose of this repository is to reproduce the original ROM. Preserve its ob
 
 Do not fix an original-game bug, remove original behavior, modernize an algorithm, strengthen validation, reorder effects, or perform speculative cleanup as part of ordinary decompilation. Report the problem and preserve the original result.
 
-Preserving the original result does not require preserving a provisional source expression that invokes undefined behavior. When a defined C expression produces the same instructions and behavior, prefer the defined expression. Do not introduce a behavioral change merely to make the source look safer.
+Preserving the original result does not require preserving a provisional source expression that invokes undefined behavior. When a defined C expression produces the same instructions and behavior, prefer the defined expression. Do not introduce a behavioral change merely to make the source look safer. If the original behavior can only be matched with a source form that actually relies on undefined behavior, keep that form only after the defined alternatives have been tested and mark it with a local `TODO` that identifies the undefined behavior and why it must remain for matching. All intentionally preserved undefined behavior must be searchable through `TODO`; do not document it only with an ordinary explanatory comment.
 
 ### Do not guess
 
@@ -203,6 +203,14 @@ Prefer defined C semantics when equivalent expressions match. Multiplication may
 
 Expression spelling can affect old compiler output. Parentheses, temporary variables, statement expressions, pointer arithmetic form, pre-increment versus post-increment, and the order in which subexpressions are written may change register allocation or instruction selection. Do not rewrite a matching expression for style without comparing the output.
 
+### 3.9 Fixed-point values
+
+Treat an integer as fixed-point only when its scale is established by consistent shifts, masks, conversion helpers, constants, field use, or arithmetic across the relevant code. Do not infer a real-number interpretation from one multiplication or shift in isolation.
+
+Preserve the verified storage width, signedness, scale, rounding, truncation, saturation or wraparound behavior, and the order of multiply, divide, add, and shift operations. Algebraically equivalent real-number formulas can differ for fixed-point integers because intermediate truncation and overflow occur at different points.
+
+Use an existing repository fixed-point type or conversion macro when it accurately represents the established format. Do not invent a new fixed-point abstraction for one uncertain value, and do not convert fixed-point state to `float` or `double` merely to make the mathematical intent clearer; the integer representation is part of the original program unless the binary proves otherwise.
+
 ## 4. Linkage and source organization
 
 ### 4.1 Hidden references and `static`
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,230 @@
...
+---
+
+## Matching
+
+**A real match is always better than a fake one.** If you write something
+non-obvious purely to make the compiler produce the right instructions, add a
+comment saying what the natural form would have been.
+
+**If it will not match, use the sanctioned escape hatch** rather than grinding.
+Put a decomp.me link and the percentage above it, as the existing ones have:
```
jiangzhengwenjz: Is it possible for an agent to create such a link?

AGENTS.md
```diff
@@ -0,0 +1,230 @@
...
+**Use `bool32` with `TRUE` and `FALSE` when a value is genuinely a boolean** — but
+only then. If you are not sure a function returns a boolean, leaving it as `u32`
+is better than guessing. Write `if (x)`, not `if (x != FALSE)`.
+
+**Write `void` explicitly** for functions that take no arguments.
+
+**Do not add casts that are not needed.**
+
+**A multi-statement macro should use a statement expression `({ ... })`**, not
+`do { ... } while (0)`. It behaves the same and usually does not change codegen.
```
jiangzhengwenjz: `do { ... } while (0)` results in binary differences in most cases, whereas `({ ... })` has also caused them a few times though less frequently; thus, using "usually" here feels a bit too broad.

And they don't behave the same. `({ ... })` is an expression (i.e. can be evaluated) but `do { ... } while (0)` is not.

AGENTS.md
```diff
@@ -0,0 +1,230 @@
+# Instructions for AI agents
+
+If you are an AI coding assistant working on this repository, read this first and
```
jiangzhengwenjz: Can you break lines semantically? This file is meant for AI agents so the current format does not help. And it will cause unnecessary line diffs every time the file gets changed.

AGENTS.md
```diff
@@ -0,0 +1,230 @@
...
+
+---
+
+## Types and constants
+
+**Use the constants that already exist.** `DISPCNT_*`, `BLDALPHA_BLEND`,
+`RGB_WHITE`, `KEYS_MASK` and the rest are in `gba/io_reg.h`, `palette.h` and
+`include/constants/`. Sound effects are named in `include/constants/songs.h`.
+
+**Use `sizeof`** instead of writing the size as a number.
```
jiangzhengwenjz: Maybe we can change this to `Use sizeof when the value genuinely means the size of the complete typed object and doing so preserves the match.`

AGENTS.md
```diff
@@ -0,0 +1,230 @@
...
+
+**Use the existing macros** rather than open-coding what they do:
+`TaskGetStructPtr`, `PlaySfx`, the `SpriteInit*` family, `Macro_*`. If a macro
+almost fits but not quite, it is usually better to add a parameter to the macro
```
jiangzhengwenjz: only when the new parameter represents a genuinely shared abstraction

AGENTS.md
```diff
@@ -0,0 +1,230 @@
...
+slightly shorter and full of tricks.
+
+**`make compare` does not check the `#else` body.** That code is not built in the
+normal build, so six green ROMs say nothing about it. If you change a file
+containing one, build it separately with `NONMATCHING` defined and confirm the
+body still compiles and has not grown.
+
+**Measure one change at a time.** If an edit touches several places, test each
+place separately. A single global search-and-replace followed by one build gives
+you one answer to several questions, and it will be the wrong answer.
```
jiangzhengwenjz: I don't see what the point is and how this is feasible.

AGENTS.md
```diff
@@ -0,0 +1,230 @@
...
+}; /* size = 0xB8 */
+```
+
+---
+
+## Types and constants
+
+**Use the constants that already exist.** `DISPCNT_*`, `BLDALPHA_BLEND`,
+`RGB_WHITE`, `KEYS_MASK` and the rest are in `gba/io_reg.h`, `palette.h` and
+`include/constants/`. Sound effects are named in `include/constants/songs.h`.
```
jiangzhengwenjz: There are other macros and other headers with constants so maybe just let the agent search in the repo?

AGENTS.md
```diff
@@ -0,0 +1,230 @@
...
+if the number of *decisions* is small — what makes a pull request reviewable is
+how many judgements a person has to check, not how many lines changed.
+
+**Before you open a pull request:**
+
+- `make compare` passes and prints **six** `OK` lines: `katam.gba`, `payload.gba`,
+  `speed_eaters.gba`, `unk_8D94B9C.gba`, `unk_8E1FE28.gba`, `unk_8E8490C.gba`.
+  Check for all six. A partial pass looks very like a full one.
+- Do not commit editor or tooling files: `.clangd`, `compile_flags.txt`, personal
+  `.gitignore` entries, scratch scripts.
```
jiangzhengwenjz: There can be other files so the examples you give are not covering everything

AGENTS.md
```diff
@@ -0,0 +1,230 @@
...
+**Write `void` explicitly** for functions that take no arguments.
+
+**Do not add casts that are not needed.**
+
+**A multi-statement macro should use a statement expression `({ ... })`**, not
+`do { ... } while (0)`. It behaves the same and usually does not change codegen.
+
+---
+
+## Formatting
```
jiangzhengwenjz: This part surely doesn't cover everything.

jiangzhengwenjz (changes requested):
And several points that I manage to think of are missing:
- Add a TODO comment when the agent find a bug or undefined behavior.
- The agent should infer types from all accesses, not from one convenient use. Try to eliminate explicit casts. For variables already declared elsewhere in the repo, they can alter them but only when necessary (like `gUnk_02038990`). 
- Do not use NODEP or MODERN for final verification.
- Also need to build the whole project with NONMATCHING defined. It won't match but at least the project must build. The doc currently only covers the file changed.
- The doc currently imposes no guidance on data decompilation.
- Learn the tricks for matching functions in the repo (like empty inline asm statements). And use if necessary.

Finally, the doc's structure is highly disorganized and very confusing; please improve it.

AGENTS.md
```diff
@@ -0,0 +1,230 @@
...
+**Give each struct a size comment** so identical structs are easy to spot:
+
+```c
+struct VertSlidingDoor {
+    struct Object2 obj2;
+    u16 unkB4;
+    s8 unkB6;
+    s8 unkB7;
+}; /* size = 0xB8 */
+```
```
soiryl-pas: Also add member offsets as comment before each member. (I'm adamant about that as it helped me find out structs multiple times)

AGENTS.md
jackmgaller: Done — rewritten with semantic line breaks throughout, one clause or sentence per line, no hard wrapping. Future edits should now produce meaningful diffs.

AGENTS.md
jackmgaller: Taken verbatim: *"Use `sizeof` when the value genuinely means the size of the complete typed object and doing so preserves the match."* Better than what I had, which implied `sizeof` is always right.

AGENTS.md
jackmgaller: Added — "only when the new parameter represents a genuinely shared abstraction, not when it is really two different things wearing one name."

AGENTS.md
jackmgaller: You're right and my sentence was wrong on both halves. Rewritten:

> A multi-statement macro should normally use a statement expression `({ ... })` rather than `do { ... } while (0)`.
> They are not equivalent: `({ ... })` is an expression and can be evaluated, `do { ... } while (0)` cannot.
> Both can change the generated code, so check the match either way.

I had claimed they "behave the same" and that `({ })` "usually" doesn't change codegen; neither is true.

AGENTS.md
jackmgaller: Replaced the list with an instruction to search:

> Before writing a literal or open-coding something, search the repository for an existing name.
> [...] Search rather than rely on any list, including this one.

A fixed list would go stale and would teach the agent to stop looking.

AGENTS.md
jackmgaller: Same fix — it now describes the category rather than enumerating:

> Only files that belong to the build are committed. Editor configuration, scratch scripts, personal ignore entries and generated files stay out. If you are unsure whether a file belongs, ask rather than commit it.

AGENTS.md
jackmgaller: Reworded so it does not read as exhaustive — it now says "Match the file you are editing. These are the rules most often corrected in review:" before the list.

AGENTS.md
jackmgaller: No, it can't — an agent has no way to create a decomp.me scratch, so asking for one was wrong. Changed to:

> If a scratch already exists for the function, link it in a comment above; do not invent a link.

That keeps the existing convention visible without asking for something the agent cannot do, and rules out the failure mode of a plausible-looking fabricated URL.

AGENTS.md
jackmgaller: Fair — I'd written it as a general rule without saying what it is for, and testing every site separately often isn't practical. Cut the standalone rule and kept only the narrow, cheap part, folded into the reporting advice:

> When you report that something does not match, show the instructions that differ rather than only saying it failed.
> Make sure you tested exactly the one change you are describing — a single global search-and-replace followed by one build gives one answer to several questions.

The concrete case was mine on #373: I told you `if (tmp) objBase = tmp;` was load-bearing, having deleted it from *both* spawn functions with one regex and reported the combined failure. It was a no-op in `sub_080B351C`, which is what you found. So the rule is not "test every site" but "do not report a combined result as a single finding". Reworded to say only that.

jackmgaller:
Rewritten. All nine inline comments and all six missing topics are addressed; replies are in the threads.

**Structure.** You were right that it was disorganised — it mixed process and reference material with overlapping sections. It now follows the order an agent needs things, one topic per section:

0. Three things that override everything else
1. Scope of one pull request
2. Writing C — naming / types / where declarations go / structs / existing constants and macros / formatting
3. Matching
4. Decompiling data
5. Flagging problems instead of fixing them
6. Verifying
7. Submitting
8. If you are unsure

**Semantic line breaks** throughout, so edits produce meaningful diffs.

**The six missing topics, all added:**

- `TODO` comments for bugs and undefined behaviour, with "do not silently correct the game's own mistakes — the build has to reproduce the original ROM, mistakes included".
- Infer types from **all** accesses, not one convenient use; try to eliminate casts, since a cast usually means a declaration is wrong; existing declarations may be changed only when necessary, and say so.
- Do not use `NODEP` or `MODERN` for final verification — those are for fast iteration only.
- Build the whole project with `NONMATCHING` defined. It will not match and is not meant to, but it must compile and link — and it is the only thing that catches a broken `#else` body, since `make compare` never builds them.
- Decompiling data: `src/data/<group>/<name>.c` with a typed named array, the `linker.ld` entry moved, structs with offset comments, `RGB()` and friends, decimal for counts and coordinates.
- The repo's matching tricks — `asm("":::"rN")`, `++x; --x;`, `x->f = x->f; // regswap`, redundant locals to move a store out of a branch — with an instruction to read the surrounding files, learn them, and comment them where used.

**Also:** struct example now carries `/* 0x00 */` offset comments on every member, per your point that it has repeatedly helped you identify structs.

Two of your comments changed my mind rather than just my wording — the decomp.me link (an agent cannot create one, so asking for it invited a fabricated URL) and the "measure one change at a time" rule (too broad and often impractical; narrowed to "do not report a combined result as a single finding").

Still documentation only.

AGENTS.md
```diff
@@ -0,0 +1,319 @@
+# Instructions for AI agents
+
+This file is for AI coding assistants working on this repository.
```
jiangzhengwenjz: I must apologize because I changed my mind. Can you change the doc so that there's one paragraph per source line, with no hard wrapping inside paragraphs?

AGENTS.md
```diff
@@ -0,0 +1,319 @@
...
+Keep the `#else` body honest and readable.
+People continue these in decomp.me scratches,
+so a body that is close and clear is worth more than one that is shorter and full of tricks.
+If a scratch already exists for the function, link it in a comment above;
+do not invent a link.
+
+When you report that something does not match,
+show the instructions that differ rather than only saying it failed.
+Make sure you tested exactly the one change you are describing —
+a single global search-and-replace followed by one build gives one answer to several questions.
```
jiangzhengwenjz: These sentences combine two separate ideas without a clear transition.
The first point is about how to report a mismatch: show the differing instructions instead of merely saying that the build failed. The next point is about experimental isolation: when investigating code generation, change one independent variable at a time so that the result can be attributed to that change. They should be separate paragraphs or separate checklist items.

The final sentence is also inaccurate as written:
> a single global search-and-replace followed by one build gives one answer to several questions

A repository-wide search-and-replace can still represent one semantic change, especially for a mechanical symbol rename. The problem is not that the edit is global; the problem is changing several independent things before performing a single build, because the resulting match or mismatch cannot then be attributed to any one of them.

jiangzhengwenjz (commented):
The doc is still disorganized and I'll let GPT generate a better document. it's impossible to continue my review at the moment.

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+
+### 4.3 Recoverable source boundaries
+
+Use linker order, symbol contiguity, shared static data, internal call structure, and cross-references to infer likely source-file ownership. Do not split or merge files merely for aesthetics, but correct an artificial boundary when the binary layout and shared implementation strongly support a different original translation unit.
+
+Keep unrelated systems in separate files. Move declarations, static data, and helper functions with their owning subsystem only when ownership is established.
+
+Do not turn a large general header into a dumping ground merely to avoid identifying ownership. Conversely, do not invent a new subsystem header for one speculative relationship.
+
+## 5. Writing C for decompilation
```
soiryl-pas, line 248: Add: Cardinal numbers to all subheadings in chapter 5, like in all other chapters

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+
+Reconsider unnecessary temporaries as well as missing ones. Recomputing a simple expression at its two natural use sites can match better than extending the lifetime of a temporary, while an aggregate or meaningful temporary can match better than repeatedly decomposing the same conceptual value. Choose between them from the instruction diff and the recovered source abstraction, not from a blanket preference for fewer or more locals.
+
+### Reuse real abstractions
+
+Do not open-code an operation when an existing repository constant, macro, helper, or typed field expresses it accurately. Search `include/constants/`, hardware register headers, subsystem headers, and nearby source before adding a literal or duplicate helper.
+
+Do not extend a macro merely to avoid a small local expression. Add a parameter only when it represents a genuinely shared abstraction, then audit and verify every affected call site.
```
soiryl-pas:
```suggestion
Do not extend or create new macros merely to avoid a small local expression. Add a parameter only when it represents a genuinely shared abstraction, then audit and verify every affected call site.
```

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+
+A GNU statement expression, a compound block, an expression macro, and `do { ... } while (0)` are not interchangeable. Preserve the form required by the macro's intended use and verify the generated code rather than changing it for style alone.
```
soiryl-pas: Maybe add: Preference for statement expressions, if all variants equally match

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+
+If a natural expression begins to match after nearby code or types are improved, remove the obsolete trick and its comment.
+
+## 6. Matching workflow
+
+### 6.1 Diagnose the mismatch before changing source
+
+Compare the actual instructions and identify the first meaningful divergence. Do not report only that a function failed to match.
+
+Classify the mismatch before choosing an experiment. Common categories include control-flow shape, branch direction, shared-tail formation, register identity, live-range length, spill or reload placement, stack-slot layout, stack-argument placement, address materialization, load or store width, signed or unsigned extension, constant materialization, alias analysis, expression ordering, helper inlining, and call ABI.
```
soiryl-pas: Add: Loop-invariant code motion (loading of constants before the loop, even though it looks like loads/calculations per iteration inside the loop body).

I feel like this happens so often that it may be emphasised, even though it obv also has its own section in 6.16

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+### 6.3 Exhaust structural explanations before compiler steering
+
+Matching tricks may be used only after ordinary reconstruction has been pursued seriously. Do not move to register clobbers, dummy operations, volatile casts, fixed-register locals, statement-expression barriers, or inline instructions merely because the first readable draft differs.
+
+Use the following escalation order. Later steps do not replace the obligation to revisit earlier ones when new evidence appears:
+
+1. Verify the function range, reference object, compiler flags, relocations, and comparison method.
+2. Recover the behavior and mirror the reference control-flow graph, including branch polarity, loop form, shared tails, early exits, and callback transitions.
+3. Correct prototypes, return types, signedness, widths, qualifiers, pointer targets, struct layouts, array strides, and hidden aggregate relationships.
+4. Recover likely source abstractions such as subobject pointers, vectors, rectangles, records, and meaningful temporaries instead of preserving an assembly-oriented decomposition into unrelated scalars.
```
soiryl-pas: Vectors and rectangles seem a bit out of place in this very general enumeration of data structures, as they're rather examples of meaningful records. Maybe just leave it at the records/structs

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+
+A trick that matches a function while structural questions remain unresolved is provisional. The pull request must explain which normal reconstructions were tested, why they were rejected, what exact mismatch remains, and why the chosen trick is the narrowest justified intervention.
+
+### 6.4 Compiler model for matching work
+
+The matching compiler identifies itself as GCC `2.9-arm-000512`. Its output is the result of several distinct stages rather than one direct translation from C syntax to Thumb instructions, so a source edit can affect a later instruction through an earlier change to RTL, control flow, liveness, pseudo-register numbering, allocation, reload, or cross-jumping.
```
soiryl-pas: Maybe add: Link to the specific agbcc repo of katam

Agents should pick it up from the INSTALL.md after reading this document anyways, but maybe it's still good to specify it lest the agent want to pick up some other gcc repo it happens to find

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+
+Assigning a constant or global address to a local creates a pseudo with a definition and live range. Directly repeating the constant or global expression gives CSE and reload different opportunities to reuse or rematerialize it. Either form can produce a pool load or a retained register; neither is guaranteed.
+
+Use a local pointer when the program logically holds the address across several operations. Use the global directly when each access is conceptually independent. Do not introduce a long-lived pointer solely to suppress a pool load unless the target and the resulting live range are verified.
+
+Pointer arithmetic and integer arithmetic are not interchangeable. Pointer addition scales by the pointed-to type and retains pointer semantics, while integer arithmetic operates on the numeric address representation. Use an integer representation for an address only when the original interface or hardware calculation genuinely requires it.
```
soiryl-pas, line 636: Maybe add: Example

The example should also highlight that different order of addition, as well as different bracket placement (through precedence of pointer arithmetic or integer arithmetic) may influence codegen

Maybe it also makes sense to write down that an integer representation of an address (if eventually deemed necessary) should have `uintptr_t` type, to make it explicit

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+```
+
+```c
+for (i = 0; i < 0x0F; ++i)
+    palette[0xE1 + i] = 0;
+```
+
+The same locations are written, but the induction variable's initial value, bound, and address constant differ. Preserve the form supported by the loop setup and compare instructions.
+
+A value loaded inside a loop may be hoisted when it is loop-invariant and alias analysis permits reuse. Moving the load into an explicit local before the loop makes the hoisting part of the source; keeping a direct memory expression leaves the decision to loop optimization and CSE. For volatile or possibly modified memory, these forms are not equivalent.
```
soiryl-pas, line 737: Maybe add: Prefer the non-explicit form without specific locals, if a lift of constants/computations through optimisations matches too.

At least I assume that in the general case where both explicit loads before and CSE based lifts match, the latter is more likely to have been written

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+
+Splitting index scaling and field offset into locals can create separate pseudos and change the order of address arithmetic. This is useful when the target holds `index * elementSize` and a fixed field offset in different registers, but the locals should represent real intermediates rather than arbitrary numbers chosen only to shuffle allocation.
+
+A `while (1)` loop and a `do { ... } while (1)` loop have different entry structure even though neither has a normal exit test. A `do` loop necessarily enters the body before the backedge; a `while` loop is represented with a header test position. Jump optimization may simplify both, but it need not produce identical block order.
+
+A small inline helper can cause an address expression to be expanded at each use after inlining rather than preserved in one explicitly reused local. This can encourage recomputation, but it is not a guaranteed anti-CSE device; inspect the RTL or final instructions before documenting that effect.
+
```
soiryl-pas, line 768: Add: if-do-while construct as for loop

Automatic decompilers like m2c or Ghidra reconstruct asm sequences that are matched with for-loops as if branches with a do-while loop within, as that very faithfully captures the asm. A separate paragraph or example about that may be valuable

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+| One hard register must be unavailable briefly | Add a narrow hard-register clobber | The named hard register is clobbered and unavailable across the asm |
+| Memory value is reused but the target reloads it | Add a narrow `memory` clobber or a semantically justified volatile access | Memory knowledge is invalidated at that point; exact reload choice remains allocator-dependent |
+| Scalar should occupy stack storage | Recover an actual array, address-taken object, aggregate, or volatile local | The object becomes addressable memory; do not invent a false type solely for frame shape |
+| Two values use the wrong registers | Change declaration order, split or reuse a temporary, or adjust one live range | Pseudo numbering, births, deaths, references, and conflicts change; the resulting assignment must be measured |
+| Add operands are reversed or a move is missing | Try three-address versus copy-then-compound-add source | The backend has both tied-output and three-register add alternatives |
+| Logical destination register is wrong | Reuse or separate the first operand around `|`, `^`, or `&` | Thumb logical patterns tie the output to one input |
```
soiryl-pas: Change: `|` Pipe is interpreted as column separator, escape

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+
+Preserve pointer semantics through relocations. Prefer symbolic pointers and function pointers over hard-coded ROM addresses when the original data is relocatable in the object file. Verify the linked values after relocation; matching unrelocated bytes with zero placeholders is not sufficient.
+
+Use explicit padding fields or byte arrays only when padding bytes belong to the recovered object. Do not absorb linker-inserted inter-object padding into the preceding struct merely to make the byte count convenient.
+
+### 9.4 Preserve order, section, and visibility
+
+Preserve the exact order of objects within each input section and the exact order of input objects in `linker.ld`. Reordering two equal-sized tables still changes every pointer or address that refers to them.
+
+Keep read-only typed data in `.rodata`, writable initialized data in `.data`, and uninitialized storage in the appropriate `NOLOAD` or `.bss` region according to the original memory map. Do not use section attributes to force an object into a desired location unless the ordinary type and declaration cannot express the verified original placement.
```
soiryl-pas, line 1019: Maybe add: Assumption that data referenced in one file only, but with wildly different rom offsets may be due to placement in .rodata and .data

However, it could also make sense to just let the agent figure this out and not constrain it beforehand on the topic of data placement

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+### 9.6 Partial `incbin` replacement
+
+When replacing part of an `incbin`, calculate the exact original start, converted length, and remaining range. Update the assembly offsets so that no byte is duplicated or omitted. Review the first bytes before the converted range and the first bytes after it to catch off-by-one and alignment mistakes.
+
+Do not leave two definitions for the same ROM range. Remove the replaced bytes from the raw source, remove obsolete linker symbols or declarations, and search for stale references to the old address-derived symbol.
+
+If the converted C object introduces relocations, compare the fully linked ROM rather than only the object payload. Inspect the relocation table when a pointer value or function address differs unexpectedly.
+
+If a raw range contains mixed typed data and opaque payload, it is acceptable to convert only the established portion. Do not invent a type for the remainder to make the conversion appear complete.
+
```
soiryl-pas, line 1060: Add: Conversion to assets of non-code form (Here or in 9.8)

Should an agent be able to generate and check in PNGs for example? With what heuristics could an agent infer that a symbol is associated with such a non-code data asset? Or should this first just be documented, but left to humans? Not sure what a good way to proceed would be, but it'd probably be good to specify some path

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+```sh
+make compare -j$(nproc)
+```
+
+Do not use `NODEP=1` or `MODERN=1` for final verification. `NODEP=1` is only an iteration shortcut when changes are limited to `.c` or `.s`; changes to other inputs can be ignored. The modern build is not the authoritative matching target.
+
+When headers, linker scripts, data, assets, build rules, generator inputs, shared declarations, or configuration changed, remove stale build products or perform a clean rebuild before trusting the result.
+
+The root `make compare` target verifies `katam.gba`; it does not by itself promise that every auxiliary multiboot ROM was rebuilt and compared. If the change can affect a multiboot subproject, shared source, shared headers, shared tools, generated data consumed by a subproject, or common build logic, run the affected root targets with `COMPARE=1` as well.
```
soiryl-pas: > it does not by itself promise that every auxiliary multiboot ROM was rebuilt and compared

Hmm, is that really the case? Looking through the Makefiles, I think when using the explicit name of a multiboot subgame as target for `make`, then setting `COMPARE=1` is necessary, but `make compare` itself populates it at the beginning and passes it to all other sub-Makefiles, so it should verify multiboot ROMs as well.

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+
+In the pull request, state:
+
+- What was inspected.
+- What the evidence establishes.
+- What competing interpretations remain possible.
+- What was deliberately left unchanged.
+- What additional evidence would resolve the uncertainty.
+
+A precise question supported by evidence is useful. A guess presented as fact creates future decompilation work and should not be committed.
```
soiryl-pas:
```suggestion
A precise question supported by evidence is useful. A guess presented as fact may impede future decompilation work and should not be committed.
```

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+
+Use a subsystem-specific header when one exists. Until a proper owner is known, place general function declarations in `functions.h` and data declarations in `data.h`, preserving the repository's ordering conventions.
+
+Put includes in the `.c` file unless the header itself requires the complete declaration. Use a forward declaration when only an incomplete type is needed and doing so matches repository style.
+
+Do not duplicate an existing declaration in another header. After changing a shared declaration, audit C, assembly, headers, data sources, linker scripts, raw pointer tables, and build configuration.
+
+### 4.3 Recoverable source boundaries
+
+Use linker order, symbol contiguity, shared static data, internal call structure, and cross-references to infer likely source-file ownership. Do not split or merge files merely for aesthetics, but correct an artificial boundary when the binary layout and shared implementation strongly support a different original translation unit.
```
soiryl-pas, line 236: Maybe add: Non-static inlines for file-boundary inference

Non-static inline functions are always emitted at the bottom of a translation unit by agbcc, so if a function is found that is inlined in other functions, this can be used to infer source boundaries

AGENTS.md
```diff
@@ -0,0 +1,1190 @@
...
+
+Check function-pointer types as carefully as data pointers. A mismatched prototype can alter argument extension, register use, stack layout, and return-value handling even when a cast suppresses the compiler warning.
+
+### 3.8 Constants and source expressions
+
+Use an existing named constant when it expresses the verified meaning exactly. Do not replace an independently meaningful protocol length, encoded field, hardware value, partial-copy size, or original table index with `sizeof` merely because the numeric values currently agree.
+
+Use `sizeof` when the value genuinely means the complete size of the typed object and the resulting code matches.
+
+Prefer defined C semantics when equivalent expressions match. Multiplication may be preferable to a signed left shift; a typed load may be preferable to aliasing through an incompatible pointer; and a range check may be preferable to an overflowing expression. Preserve the ROM result, not accidental undefined behavior in a provisional decompilation.
```
soiryl-pas, line 202: Add: If actual UB is needed, add a comment explaining it

While also being addressed in chapter 8, it would very specifically help with #64

soiryl-pas (changes requested):
I annotated a few points that may be worthwhile to change.
As for general things that are missing, the only thing I can think of is perhaps some notes to the handling of fixed-point real numbers. Not sure what would be valueable to write though, as they aren't really of big importance in our code base anyways.

AGENTS.md
jiangzhengwenjz: GPT:
>I agree that vectors and rectangles are structurally just records/structs, but I would prefer to keep them as concrete examples rather than remove them entirely.
One recurring matching failure mode is writing provisional C too close to the assembly and decomposing a conceptual aggregate into unrelated scalar temporaries. Recovering an existing vector-like or rectangle-like type can restore the likely source-level abstraction and materially change the generated code.
I’ll reword this to make the relationship clearer, e.g. “meaningful records/structs, such as existing vector or rectangle types when supported by the accesses.”

And I think it makes sense - it knows agents better than us

AGENTS.md
jiangzhengwenjz: GPT:
> That's a useful observation, and I agree that this should be included as an additional clue for recovering translation-unit boundaries.
I checked the relevant agbcc behavior so that I can phrase the rule more precisely. Top-level functions that agbcc considers inlinable can have their output deferred until end-of-file processing, so an out-of-line copy of a non-static inline function will commonly appear after the normally emitted functions from the same translation unit. That makes its placement useful evidence when reconstructing source-file boundaries.
There are cases where an `inline` function is rejected by the inliner—for example, agbcc explicitly rejects variadic functions—so I don't want to phrase the rule as an unconditional “all non-static inline functions are always emitted at the bottom.”
I'll add this to the source-boundary section as a concrete heuristic: when a non-static inline function is emitted late, treat that as supporting evidence that it belongs to the preceding translation unit, and check the function's inline eligibility and surrounding linker/symbol evidence before relying on it as a boundary marker.

So GPT has found an exception to your argument - but we'll work on this in the next commit.

AGENTS.md
jiangzhengwenjz: GPT:
> Yes, I think this should be documented more specifically.
There is evidence that the still-raw region currently represented by `data/data_2.s` is, at least in substantial part and likely as a whole, the original `.data` region rather than ordinary `.rodata`.
The current boundaries are suggestive: `data_2.s` starts at `0x08D60980`, immediately after the final small library `.rodata` fragment and `impure.o(.rodata)`, and ends at `0x08D6DD04`, exactly where the embedded multiboot data begins. We also already have matching evidence from code referencing a symbol in this region that the symbol needs `.data`-style placement/declaration to reproduce the original code.
So rather than only saying that ROM distance may result from `.rodata`/`.data` separation, I'll add a repository-specific note that agents decompiling the `0x08D60980–0x08D6DD04` region should actively test whether recovered objects belong in `.data` and reconstruct the linker placement accordingly, instead of assuming that the current raw `.rodata` representation reflects the original source section.
I would still treat the range as a strong working hypothesis rather than claiming every object in it is already proven to be `.data`; individual recovered objects should be checked against their declarations, references, generated sections, and matching behavior.

The new doc will be based on the inferred range.

AGENTS.md
soiryl-pas: The rephrasing sounds good

AGENTS.md
soiryl-pas: Ah, excellent
Variadic nonstatic inlines not being emitted, that I didn't really know

soiryl-pas (approved):
I went over the diff to check whether all suggestions have been implemented, and it looks good to go to me.

jiangzhengwenjz (approved):

PR #382, opened 2026-08-08
update calcrom.pl

jiangzhengwenjz:
AI-Assisted-by: ChatGPT (GPT-5.6 Thinking)

## Summary

This updates `calcrom.pl` to make its statistics more literal and easier to interpret. The main goal is to separate **where bytes are currently linked from** from **what work still remains to be decompiled**, while keeping the existing `src`/`asm` view useful.

### Count all mapped code by origin

`total bytes of code` now includes every non-zero `.text` input section found in the linker map, not only objects under `src/` and `asm/`. Archive members such as `libc.a` and `libgcc.a` therefore contribute to the total as well.

**Rationale:** the previous total was not actually the total amount of code in the ROM. The report now treats `bytes of code in/from ...` as a mechanical provenance breakdown, so the wording can be read literally.

Zero-sized input sections are discarded while parsing and never create an origin category.

### Split assembly provenance from decompilation progress

`bytes of code in asm` continues to include all assembly-backed code. It is then broken down into three mutually exclusive categories:

- **bytes of code remaining to be decompiled** — ordinary assembly that still needs a C decompilation;
- **bytes in `asm/nonmatching` functions** — functions that already have C implementations but still use assembly fallback because they do not match;
- **bytes excluded from decompilation tracking** — intentionally retained assembly such as runtime/handwritten assembly already excluded by the script.

The number of functions represented by `asm/nonmatching/*.inc` is also reported.

**Rationale:** being written in assembly and still needing decompilation are not the same thing. `asm/nonmatching` code has already been decompiled, while intentionally retained assembly may never be decompiled at all. Keeping these categories separate makes both the physical `asm` total and the remaining workload meaningful.

The three asm subcategories are indented beneath the asm total, while all top-level code origins are indented beneath the overall code total.

### Count ROM data mechanically by origin

Data statistics now include every non-zero `.rodata` and initialized `.data` input section linked into the ROM and group those bytes by origin. This includes sources such as `sound` and archive members in addition to `src` and `data`.

The previous `Data porting to C is 100% complete` interpretation is removed.

**Rationale:** the location of data in `src/`, `data/`, `sound/`, or an archive is an objective property of the build, but it is not a reliable measure of whether that data has been fully decompiled. In particular, data can already have had its raw incbin removed while still remaining in an assembly data file. The script now reports provenance without assigning an unsupported completion meaning to it.

The origin breakdown is indented beneath `total bytes of data` to make clear that those categories form the total.

### Restrict incbin statistics to direct baserom dependencies

The incbin scan now counts only `.incbin` directives that directly reference a `baserom*.gba` file. Generated/preprocessed directories such as `build`, `.git`, and `tools` are skipped.

The baserom-incbin line also reports its percentage of total ROM data and is kept outside the indented origin breakdown.

**Rationale:** an `.incbin` is not inherently undecompiled data. Generated binaries, multiboot payloads, graphics, audio, and other intentional binary assets may legitimately remain incbin inputs even when their source is fully reconstructed. Direct baserom incbins are the mechanically identifiable case that still depends on bytes copied from the original ROM.

This statistic is intentionally a separate dimension from the origin breakdown: baserom incbins may themselves live under `data/`, `src/`, or another origin, so the percentages are not meant to sum together.

### Parse nonmatching symbol sizes once

The script now invokes `nm -S` once and records only exact `T`/`t` symbol matches for functions represented by `asm/nonmatching/*.inc`.

**Rationale:** the previous implementation invoked `nm | grep | awk` once per function, which was slower and allowed substring matches. Limiting the lookup to the exact nonmatching function names also avoids failures caused by unrelated duplicate local/library symbols in the ELF.

### Add consistency checks

The script verifies that:

- raw asm bytes equal `remaining to be decompiled + excluded`;
- adjusted asm bytes equal `remaining to be decompiled + nonmatching + excluded`;
- nonmatching bytes do not exceed the bytes initially attributed to `src`.

**Rationale:** these categories are intended to be an exact partition. If future linker/layout changes cause the parser or classification rules to miss something, the script should fail visibly rather than print internally inconsistent progress numbers.

### Fix 4-byte size rounding

The size-alignment correction now uses `% 4` instead of `% 3`.

**Rationale:** the surrounding condition checks 4-byte alignment (`$size & 3`), so `% 3` was inconsistent and could round values to the wrong size.

## Output semantics

The resulting report now has two distinct kinds of information:

- **Code/data provenance:** where ROM bytes come from in the linked image.
- **Decompilation status:** only the explicitly identified assembly categories that represent remaining work or nonmatching C implementations.

This intentionally avoids deriving a general data-decompilation percentage from directory placement or from the presence of arbitrary `.incbin` directives.

## Testing

- `perl -c calcrom.pl`
- Synthetic map and controlled `nm` fixtures covering:
  - `src`, `asm`, excluded asm, and `asm/nonmatching`;
  - archive `.text`;
  - `.rodata` and initialized `.data`;
  - `sound` and archive data;
  - zero-sized input sections;
  - direct baserom incbins and generated-binary incbins;
  - unrelated duplicate symbols in `nm -S`.
- Verified that nonmatching bytes are not counted as code remaining to be decompiled.
- Verified that the three asm subcategories sum exactly to the reported asm total.
- Verified that zero-byte input sections never create statistical origin categories.
- Verified that baserom incbins under generated/preprocessed directories are not counted twice.
- Verified that the script rejects inconsistent classifications rather than silently reporting them.

The script should additionally be run against a normal clean project build before merge so the final output can be checked against the real `katam.map` and `katam.elf`.

### DIFF
--- a/calcrom.pl
+++ b/calcrom.pl
@@ -38,8 +38,6 @@ my $excluded_decomp_bytes = 0;
 sub object_origin
 {
     my ($object) = @_;
-    $object =~ s{\\}{/}g;
-
     # Archive members are reported by ld as path/to/libfoo.a(member.o).
     return $1 if ($object =~ m{([^/]+\.a)\([^)]*\)\s*$});
 
@@ -74,10 +72,7 @@ while (my $line = <$file>)
 
             if ($origin eq 'asm')
             {
-                my $basename = $object;
-                $basename =~ s{\\}{/}g;
-                $basename =~ s{.*/}{};
-                $basename =~ s{\.o$}{};
+                my $basename = basename($object, ".o");
 
                 if ($decomp_excluded_asm{$basename})
                 {
@@ -127,18 +122,54 @@ my $nm_sizes_output;
     or die "ERROR: Error while getting symbol sizes: $?";
 
 my %symbol_sizes = ();
+my $total_syms = 0;
+my $undocumented = 0;
+my $previous_filtered_symbol;
+
 foreach my $line (split /\n/, $nm_sizes_output)
 {
-    if ($line =~ /^\s*[0-9a-fA-F]+\s+([0-9a-fA-F]+)\s+([Tt])\s+(\S+)\s*$/)
+    my ($size, $type, $name);
+
+    # nm -S does not print a size for every defined symbol. Accept both
+    # "address size type name" and "address type name" so that documentation
+    # counts preserve the behavior of the previous plain-nm pipeline.
+    if ($line =~ /^\s*[0-9a-fA-F]+\s+([0-9a-fA-F]+)\s+(\S)\s+(\S+)\s*$/)
     {
-        my $size = hex($1);
-        my $name = $3;
+        $size = hex($1);
+        $type = $2;
+        $name = $3;
+    }
+    elsif ($line =~ /^\s*[0-9a-fA-F]+\s+(\S)\s+(\S+)\s*$/)
+    {
+        $type = $1;
+        $name = $2;
+    }
+    else
+    {
+        next;
+    }
 
-        # nm can legitimately contain duplicate local/library symbol names.
-        # We only need sizes for asm/nonmatching functions, so ignore every
-        # other symbol rather than imposing a repository-wide uniqueness rule.
-        next unless exists $nonmatchings{$name};
+    # Preserve the previous documentation-counting semantics:
+    #   awk '{print $3}' | grep '^[^_].\{4\}' | uniq
+    # "uniq" only removes adjacent duplicates after filtering, so keep track
+    # of the previous accepted name rather than globally deduplicating symbols.
+    if ($name !~ /^_/ and length($name) >= 5)
+    {
+        if (!defined($previous_filtered_symbol) or $name ne $previous_filtered_symbol)
+        {
+            $total_syms++;
+            $undocumented++
+                if ($name =~ /(?:[Uu]nk_[0-9a-fA-F]*|sub_[0-9a-fA-F]*)/);
+            $previous_filtered_symbol = $name;
+        }
+    }
 
+    # Nonmatching function sizes need exact text-symbol matches with an
+    # available size field.
+    if (defined($size)
+        and ($type eq 'T' or $type eq 't')
+        and exists $nonmatchings{$name})
+    {
         if (exists $symbol_sizes{$name} and $symbol_sizes{$name} != $size)
         {
             die "ERROR: Nonmatching symbol '$name' has conflicting sizes in nm output.\n";
@@ -194,31 +225,9 @@ else
 $adjusted_asm_bytes == $remaining_decomp_bytes + $nonmatching_bytesum + $excluded_decomp_bytes
     or die "ERROR: adjusted asm classification is inconsistent.\n";
 
-# Note that the grep filters out all branch labels. It also requires a minimum
-# line length of 5, to filter out a ton of generated symbols (like AcCn). No
-# settings to nm seem to remove these symbols. Finally, nm prints out a separate
-# entry for whenever a name appears in a file, not just where it's defined. uniq
-# removes all the duplicate entries.
-my $base_cmd = "nm \"$elffname\" | awk '{print \$3}' | grep '^[^_].\\{4\\}' | uniq";
-# This looks for Unknown_, Unknown_, or sub_, followed by just numbers. Note that
-# it matches even if stuff precedes the unknown, like sUnknown/gUnknown.
-my $undoc_cmd = "grep '[Uu]nk_[0-9a-fA-F]*\\|sub_[0-9a-fA-F]*'";
-my $count_cmd = "wc -l";
-
-my $total_syms_as_string;
-(run (
-    command => "$base_cmd | $count_cmd",
-    buffer => \$total_syms_as_string,
-    timeout => 60
-))
-    or die "ERROR: Error while getting all symbols: $?";
-my $undocumented_as_string;
-(run (
-    command => "$base_cmd | $undoc_cmd | $count_cmd",
-    buffer => \$undocumented_as_string,
-    timeout => 60
-))
-    or die "ERROR: Error while filtering for undocumented symbols: $?";
+# Symbol documentation counts are derived from the same nm -S output above.
+# This preserves the previous filtering rules while avoiding two more nm
+# invocations and the accompanying awk/grep/uniq/wc pipelines.
 
 # Only direct references to a baserom count as remaining incbins. Generated
 # binaries and intentional binary assets may legitimately continue to use
@@ -236,7 +245,6 @@ sub parse_integer
 sub is_baserom_path
 {
     my ($path) = @_;
-    $path =~ s{\\}{/}g;
     return basename($path) =~ /^baserom(?:_[^\/]*)?\.gba$/i;
 }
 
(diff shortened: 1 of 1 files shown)


calcrom.pl
```diff
@@ -18,108 +19,199 @@
...
-            elsif ($dir eq 'asm')
+            $code_by_origin{$origin} += $size;
+
+            if ($origin eq 'asm')
             {
-                if (!($basename =~ /(crt0|libagbsyscall|multi_sio_asm|m4a_asm)/))
+                my $basename = $object;
+                $basename =~ s{\\}{/}g;
+                $basename =~ s{.*/}{};
+                $basename =~ s{\.o$}{};
```
soiryl-pas: As I understand it, these substitutions extract the basename of a file (e.g. `asm/crt0.o` -> `crt0`)
However, File::Basename::basename is imported anyways, so it may be equivalent to just use that function with `.o` as omitted suffix. Not totally sure though

calcrom.pl
```diff
@@ -18,108 +19,199 @@
...
-# You'd expect this to take a while, because of uniq. It runs in under a second,
-# though. Uniq is pretty fast!
-my $base_cmd = "nm $elffname | awk '{print \$3}' | grep '^[^_].\\{4\\}' | uniq";
+if ($adjusted_asm_bytes > 0)
+{
+    $code_by_origin{'asm'} = $adjusted_asm_bytes;
+}
+else
+{
+    delete $code_by_origin{'asm'};
```
soiryl-pas, line 222: This looks like it would leave out a report of asm when no code is left in asm. Is that what we want? I'd find `0 bytes of code in asm` to be a wonderful message actually :D

calcrom.pl
```diff
@@ -128,68 +220,135 @@
 ))
```
soiryl-pas: I feel like these two command invocs should be able to be replaced by filtering `$nm_sizes_output` adequately, which holds the whole output of the `nm -S` invocation from before. Could be nice for performance

calcrom.pl
```diff
@@ -128,68 +220,135 @@
...
+
+            return unless ($File::Find::name =~ /\.(?:s|inc)$/);
+            open(my $incfile, '<', $File::Find::name)
+                or die "ERROR: could not open file '$File::Find::name'.\n";
+            while (my $line = <$incfile>)
+            {
+                next unless ($line =~ /^\s*\.incbin\s*"([^"]+)"(.*)$/);
+                my $path = $1;
+                my $args = $2;
+                next unless is_baserom_path($path);
```
soiryl-pas, line 274: I must admit, I'm not really sure what this incbin count part does (until line 281). So perhaps inspect it even further experimentally

soiryl-pas (changes requested):
I tried to understand what the program does and it should be alright, with some comments.
I've also seen that it oftentimes rewrites Windows-backslash paths to Linux-forward-slash-paths (`s{\\}{/}g`), which seems unnecessary for our linux-only repo, but it wouldn't hurt I guess.

calcrom.pl
jiangzhengwenjz: GPT:
> Yes, that's equivalent here and cleaner. Since `basename` is already imported and we only need to strip the `.o` suffix, I'll replace the manual substitutions with `basename($object, ".o")`.


👍

calcrom.pl
jiangzhengwenjz: GPT:
> Agreed. `$nm_sizes_output` already contains the information needed for the documentation counts, so there is no reason to invoke `nm` again twice through separate shell pipelines.
I'll reuse that output and reproduce the existing symbol-name filtering/deduplication in Perl, while preserving the current semantics of the documented/undocumented symbol counts.

calcrom.pl
jiangzhengwenjz: GPT:
> This part intentionally measures only direct dependencies on the original ROM, rather than all uses of `.incbin`.
It recursively scans source `.s`/`.inc` files, ignores generated/repository-internal directories such as `build`, `.git`, and `tools`, and only counts directives of the form:
```asm
.incbin "baserom*.gba", offset, size
```
> The explicit sizes are summed to produce the baserom-incbin byte count.
The distinction is intentional because `.incbin` itself does not imply undecompiled data. For example, a generated multiboot binary or an intentional binary asset may legitimately remain an `.incbin` even when its source is fully reconstructed. A direct slice of `baserom*.gba`, on the other hand, is something we can mechanically identify as still depending on the original ROM.
I tested this both with synthetic cases and on the current project tree. The current build reports 29,337 direct baserom incbins totaling 9,868,500 bytes, while generated/non-baserom incbins are excluded. The parser also rejects baserom incbins without an explicit offset and size rather than silently guessing their contribution.

calcrom.pl
jiangzhengwenjz: GPT:
> This block only controls whether the top-level `asm` provenance category exists after moving nonmatching fallback bytes from `src` to `asm`.
It does not suppress the more interesting completion case: `0 bytes of code remaining to be decompiled`. That value is tracked separately and will still be printed as zero as long as any asm-origin code remains, e.g. intentionally retained assembly or nonmatching fallbacks.
`adjusted_asm_bytes` itself only reaches zero when there is literally no asm-origin code and no `asm/nonmatching` fallback code at all. In that case I think omitting the zero-byte provenance category is consistent with the rest of the report, since zero-sized origins are intentionally not represented.

jiangzhengwenjz:
In case you can't do a local test, the current output is:
```
$ ./calcrom.pl katam.map --verbose
1424856 total bytes of code
    1268712 bytes of code in src (89.0414%)
    145256 bytes of code in asm (10.1944%)
        138376 bytes of code remaining to be decompiled

BREAKDOWN
    1152 bytes in asm/code_08020220.s
    1504 bytes in asm/multi_sio_08158934.s
    1652 bytes in asm/boss_challenge_door.s
    1872 bytes in asm/object_spawner.s
    1968 bytes in asm/code_08025650.s
    2420 bytes in asm/code_0802F8D8.s
    2520 bytes in asm/kirby_cannon.s
    2644 bytes in asm/big_small_switch.s
    3372 bytes in asm/code_0806F734.s
    3504 bytes in asm/shotzo.s
    3596 bytes in asm/sub_0800F044.s
    3640 bytes in asm/leap.s
    3664 bytes in asm/cookin.s
    4076 bytes in asm/hot_head.s
    4140 bytes in asm/star_stone_block.s
    4472 bytes in asm/prank.s
    4956 bytes in asm/code_0802E57C.s
    6260 bytes in asm/large_star_stone_block.s
    7392 bytes in asm/sprite.s
    8740 bytes in asm/flamer.s
    12500 bytes in asm/code_0802B4A8.s
    12656 bytes in asm/cutscene_trigger.s
    39676 bytes in asm/code_08032E98.s

NONMATCHING
    2868 bytes in asm/nonmatching/sub_08153184.inc

41547 total symbols
    6781 symbols documented (16.3213%)
    34766 symbols undocumented (83.6787%)

14512864 total bytes of data

```

jiangzhengwenjz:
And I asked GPT to explain the number difference between this output and the one produced by the original one:

### Why the reported data total increases

The reported `total bytes of data` is expected to increase compared with the previous script.

Previously, the data total was effectively limited to `.rodata` attributed to `src/` and `data/`. Other ROM-resident data origins were not included in the denominator, and initialized `.data` sections were not counted either. The older parser was also limited to simple section names and did not reliably cover subsection names such as `.rodata.*` or `.data.*`.

The new implementation instead counts every non-zero ROM-resident `.rodata`, `.rodata.*`, `.data`, and `.data.*` input section found in the map and then classifies those bytes by origin.

On the current build, the difference is directly visible in the new breakdown:

```text
125048 bytes of data in sound
948 bytes of data from libc.a
```

These 125,996 bytes were previously absent from the reported data total. They now contribute to it because they are real data bytes linked into the ROM.

This change is therefore not intended to represent a regression in data decompilation progress. It corrects the meaning of `total bytes of data`: the total now reflects the mapped ROM data being reported, rather than only the subset originating from `src/` and `data/`.

The separate 4-byte rounding fix may also change individual section accounting slightly, but it is not the main reason for the increase in the reported total.

calcrom.pl
soiryl-pas: Yes, I think that makes sense, then I think this snippet should work fine

calcrom.pl
soiryl-pas: Ah, well, we'll always have intentionally retained assembly (bc of some libs and crt0 and stuff), so the deletion branch with no asm at all would never be hit in the first place for this repo. We can still leave it in there though for consistency with the rest of the reportings

soiryl-pas (approved):
The reasoning in the pending comments is valid and from what I can see the reporting is done correctly, so I think it's ready to be merged.

PR #384, opened 2026-08-08
Name the Kirby water/swim state machine

jackmgaller:
Names the Kirby water/swim state machine in `src/kirby.c`: the water-boundary pair `KirbyEnterWater`/`KirbyLeaveWater`, the shared water-movement entry `KirbyStartWaterMovement`, the free-swim state `KirbySwim`, and the grounded underwater pair `KirbyStartWaterIdle`/`KirbyWaterIdle` and `KirbyStartWaterWalk`/`KirbyWaterWalk`.

Rebased onto current master. Purely textual — all six ROMs match from a clean build (`make compare`). 4 files, 309 insertions / 309 deletions.

This work was AI-assisted (Claude); the commit carries the co-author trailer.

**Changes in this revision (review round 2):**

- `KirbyStartSwim` → `KirbyStartWaterMovement`, per the review comment: the function's first branch dispatches to the grounded anim-56 state and only the other path enters `KirbySwim`, so the old name described one of two paths. Details and a full caller census in the section below.
- The adjacent grounded states (anim 56/57) are now named as well — the review's other suggestion — following the `StartIdle`/`Idle`/`StartWalk`/`Walk` convention from the merged Bonkers states.
- Rebased onto `ff38f11`.

**Changes from the first revision, per the recheck:**

- The first revision renamed `sub_0805BC78` → `KirbyIsInWater` while this description claimed it was deliberately left unnamed. That rename is reverted — it was undisclosed and also wrong: the function reads the collision attribute at Kirby's tile and, when bit `0x4000` is set (or the `unk62 & 0x10` latch is), calls `sub_0805545C`. The same `unk58 |= 0x4000; unk62 |= 0x10` pair is set by the door-entry path at `kirby.c:6854` under `(attr & 0xF04000) == 0x104000`, its consumer `sub_080552A8` selects a destination room and plays `SE_KIRBY_MIRROR_TO_HUB`/`SE_KIRBY_MIRROR_STANDARD`, and the handler at `kirby.c:12848` relocates Kirby through the doors-table entry. So attribute bit `0x4000` marks door/warp tiles, not water, and the function stays `sub_0805BC78`.
- "Always called immediately before the swim entry" is corrected to a measured claim: 111 of the 124 `KirbyEnterWater` call sites are immediately followed by `KirbyStartWaterMovement`. The exceptions are the in-water branches of states that keep their own update — UFO (`kirby.c:10048`, which sets `unk58 |= 2` itself), the fly/float states (`:13266`, `:13321`, `:13436`, latching `flyTimer = 1`), and the handlers at `:16434`/`:16486` (which register `sub_0805EF18`/`sub_0805F210`). Each exception has a matching `KirbyLeaveWater` on its exit edge, which is the evidence that the Enter/Leave pair marks the water boundary independent of the swim state.

---

## Manifest

**Base:** `ff38f11` (origin/master). Line numbers below were taken at `9724813`; the two commits since are renames elsewhere in the files and do not shift them.

| old | new |
|---|---|
| `sub_0805BA58` | `KirbyEnterWater` |
| `sub_08059810` | `KirbyStartWaterMovement` |
| `sub_080598E8` | `KirbySwim` |
| `sub_0805BBB4` | `KirbyLeaveWater` |
| `sub_08058628` | `KirbyStartWaterIdle` |
| `sub_08058714` | `KirbyWaterIdle` |
| `sub_08058F3C` | `KirbyStartWaterWalk` |
| `sub_08059008` | `KirbyWaterWalk` |

**The water bit:** `ObjectBase::unk58` is assigned the raw collision attribute at the object's tile every physics step (`code_0806F780.c:794`, `a1->unk58 = gCollisionAttributes[a1->unk57]`), so `unk58 & 2` is collision-attribute bit 1 at Kirby's position. The "water" reading comes from the behavior gated on it, below, not from a sound effect alone.

**The ground bit:** `unk62 & 4` distinguishes the two in-water regimes — grounded (standing/walking on the bottom, anims 56/57) versus free swimming (anim 59). Both grounded updates transition to `KirbyStartWaterMovement` the moment `unk62 & 4` clears.

### `sub_0805BA58` → `KirbyEnterWater` (124 references)

One-shot on crossing into water: creates splash effect `0x296` + `PlaySfx(SE_WATER_SPLASH)`, switches the shadow sprite to palette `0xE`, clamps `xspeed` to ±0x13C, cancels upward `yspeed` and quarters downward `yspeed` — entry damping to in-water limits. Called from the `unk58 & 2` branch of every land/air state dispatch (111 sites immediately ahead of `KirbyStartWaterMovement`) and from the non-swim in-water states listed above.

### `sub_08059810` → `KirbyStartWaterMovement` (122 references)

The shared entry that routes Kirby into the correct in-water state: grounded (`unk62 & 4`) → `KirbyStartWaterIdle`; otherwise sets animation 59, resets the state counter, sets the in-water shadow palette, clears the ground/ability flag groups, and registers `KirbySwim` as the per-frame update (`unk78 = KirbySwim`).

Caller census, to pin the "water" claim: of 117 call sites in `kirby.c`, 113 sit directly under an `unk58 & 2` check — the bulk being the `unk58 & 2` arm of the state-dispatch macro `Macro_0803FF64_6` (`kirby.c:7000`, 99 uses) and the paired `KirbyEnterWater` blocks — and the remaining 4 are inside `KirbyWaterIdle`/`KirbyWaterWalk` themselves, which run only while `unk58 & 2` holds. Both `dark_meta_knight.c` sites are gated on `unk58 & 2`, and the one assembly caller (`asm/cutscene_trigger.s:1740`) tests `[r2, #0x58] & 2` before the `bl`. All 120 call sites are in-water contexts.

### `sub_080598E8` → `KirbySwim` (3 references)

The registered update for free swimming. While `unk58 & 2` holds: swim-stroke input plays `SE_KIRBY_SWIM_1`/`SE_KIRBY_SWIM_2`, applies the rise/sink speed clamps. When `unk58 & 2` drops, hands off through `KirbyLeaveWater` and re-enters the fall/land states.

### `sub_0805BBB4` → `KirbyLeaveWater` (40 references)

One-shot on leaving water: splash effect variant (`0x296, 1`), `PlaySfx(SE_WATER_SPLASH)`, restores the shadow sprite's land palette. Called on the `!(unk58 & 2)` exit edge of `KirbySwim` and of every non-swim in-water state that called `KirbyEnterWater` (the UFO path keys the same transition off the palette latch, `kirby.c:10051`).

### `sub_08058628` / `sub_08058714` → `KirbyStartWaterIdle` / `KirbyWaterIdle` (9 / 3 references)

The grounded underwater idle state, animation 56. Entry zeroes `idleTimer`/`flyTimer` and the walk-cycle bytes, registers `KirbyWaterIdle`, and reroutes to the walk state immediately if `xspeed != 0`. The update: leaves water when `unk58 & 2` clears; transitions to `KirbyStartWaterWalk` on left/right input (`movementState & 0x20`/`0x10`) unless blocked by a wall (`unk62 & 1`); pushes off into `KirbyStartWaterMovement` on jump input (`unk11A & 1`, with `SE_KIRBY_SWIM_1` and an upward impulse) or when ground contact (`unk62 & 4`) is lost.

### `sub_08058F3C` / `sub_08059008` → `KirbyStartWaterWalk` / `KirbyWaterWalk` (5 / 3 references)

The grounded underwater walk state, animation 57. Entry and update both call `sub_0805B644`, which has no other callers repo-wide, so it is these two states' private movement routine (`sub_0805B644` is left unnamed in this PR). The update shares `KirbyWaterIdle`'s leave-water / push-off / ground-loss transitions and zeroes `xspeed` on wall contact (`unk62 & 1`).

### DIFF
--- a/src/kirby.c
+++ b/src/kirby.c
@@ -136,11 +136,11 @@ void sub_080579F4(struct Kirby *);
 void sub_08057AD0(struct Kirby *);
 void sub_08057E08(struct Kirby *);
 void sub_08058158(struct Kirby *);
-void sub_08058628(struct Kirby *);
-void sub_08058714(struct Kirby *);
-void sub_08058F3C(struct Kirby *);
-void sub_08059008(struct Kirby *);
-void sub_080598E8(struct Kirby *);
+void KirbyStartWaterIdle(struct Kirby *);
+void KirbyWaterIdle(struct Kirby *);
+void KirbyStartWaterWalk(struct Kirby *);
+void KirbyWaterWalk(struct Kirby *);
+void KirbySwim(struct Kirby *);
 void sub_0805A4EC(struct Kirby *);
 void sub_0805A64C(struct Kirby *);
 void sub_0805A838(struct Kirby *);
@@ -161,8 +161,8 @@ void sub_0805B644(struct Kirby *);
 void sub_0805B6BC(struct Kirby *);
 void sub_0805B8B8(struct Kirby *);
 void sub_0805B988(struct Kirby *);
-void sub_0805BA58(struct Kirby *);
-void sub_0805BBB4(struct Kirby *);
+void KirbyEnterWater(struct Kirby *);
+void KirbyLeaveWater(struct Kirby *);
 bool8 sub_0805BC78(struct Kirby *);
 void sub_0805BE80(struct Kirby *);
 void sub_0805BEE8(struct Kirby *);
@@ -7006,7 +7006,7 @@ void sub_0803FE74(struct Kirby *kirby)
     else if ((kirby)->ability == KIRBY_ABILITY_CUPID && (kirby)->base.base.base.flags & 0x40) \
         sub_08047EF0(kirby); \
     else if ((kirby)->base.base.base.unk58 & 2) \
-        sub_08059810(kirby); \
+        KirbyStartWaterMovement(kirby); \
     else if ((kirby)->base.base.base.flags & 0x60) \
         sub_08044EA8(kirby); \
     else \
@@ -7072,8 +7072,8 @@ void sub_0803FF64(struct Kirby *kirby)
     kirby->base.base.base.flags |= 4;
     if (kirby->base.base.base.unk58 & 2)
     {
-        sub_0805BA58(kirby);
-        sub_08059810(kirby);
+        KirbyEnterWater(kirby);
+        KirbyStartWaterMovement(kirby);
         return;
     }
     else
@@ -7230,8 +7230,8 @@ void sub_08040930(struct Kirby *kirby)
     kirby->base.base.base.flags |= 4;
     if (kirby->base.base.base.unk58 & 2)
     {
-        sub_0805BA58(kirby);
-        sub_08059810(kirby);
+        KirbyEnterWater(kirby);
+        KirbyStartWaterMovement(kirby);
         return;
     }
     if (!(kirby->base.base.base.unk62 & 4))
@@ -7362,8 +7362,8 @@ void sub_080412AC(struct Kirby *kirby)
     Macro_0803FF64_1(kirby);
     if (kirby->base.base.base.unk58 & 2)
     {
-        sub_0805BA58(kirby);
-        sub_08059810(kirby);
+        KirbyEnterWater(kirby);
+        KirbyStartWaterMovement(kirby);
         return;
     }
     if (sub_0805BC78(kirby)) return;
@@ -7488,8 +7488,8 @@ void sub_08041C50(struct Kirby *kirby)
     Macro_0803FF64_1(kirby);
     if (kirby->base.base.base.unk58 & 2)
     {
-        sub_0805BA58(kirby);
-        sub_08059810(kirby);
+        KirbyEnterWater(kirby);
+        KirbyStartWaterMovement(kirby);
         return;
     }
     if (sub_0805BC78(kirby)) return;
@@ -7604,8 +7604,8 @@ void sub_080425F0(struct Kirby *kirby)
     kirby->base.base.base.flags |= 4;
     if (kirby->base.base.base.unk58 & 2)
     {
-        sub_0805BA58(kirby);
-        sub_08059810(kirby);
+        KirbyEnterWater(kirby);
+        KirbyStartWaterMovement(kirby);
         return;
     }
     if (!(kirby->base.base.base.unk62 & 4))
@@ -7701,8 +7701,8 @@ void sub_08042E28(struct Kirby *kirby)
     kirby->base.base.base.flags |= 4;
     if (kirby->base.base.base.unk58 & 2)
     {
-        sub_0805BA58(kirby);
-        sub_08059810(kirby);
+        KirbyEnterWater(kirby);
+        KirbyStartWaterMovement(kirby);
         return;
     }
     Macro_0803FF64_3(kirby);
@@ -7882,8 +7882,8 @@ void sub_080435F8(struct Kirby *kirby)
     Macro_0803FF64_1(kirby);
     if (kirby->base.base.base.unk58 & 2)
     {
-        sub_0805BA58(kirby);
-        sub_08059810(kirby);
+        KirbyEnterWater(kirby);
+        KirbyStartWaterMovement(kirby);
         return;
     }
     if (sub_0805BC78(kirby)) return;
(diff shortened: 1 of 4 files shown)


jackmgaller:
Rechecked per your request, against a fresh reading of the code:

- The previous revision had renamed `sub_0805BC78` → `KirbyIsInWater` while the description claimed the function was deliberately left unnamed. That rename is reverted — beyond being undisclosed, it was wrong: the `unk58 |= 0x4000; unk62 |= 0x10` pair it sets is the door/warp-transition latch (same pair set by the door-entry path at `kirby.c:6854`; consumers `sub_080552A8` and the relocation handler at `:12848` route Kirby to a destination room and play the mirror-travel sounds), so collision-attribute `0x4000` marks warp tiles, not water.
- The "always called immediately before KirbyStartSwim" claim is corrected to the measured 111 of 122, with the exceptions enumerated (UFO/fly/float in-water paths that keep their own updates and pair with `KirbyLeaveWater` on exit).
- Rebased onto master; the commit now carries the co-author trailer.

The four remaining renames and their evidence are in the updated description. Ready for review.

src/kirby.c
```diff
@@ -13891,7 +13891,7 @@ void sub_08059008(struct Kirby *kirby)
         sub_08058628(kirby);
 }
 
-void sub_08059810(struct Kirby *kirby)
+void KirbyStartSwim(struct Kirby *kirby)
```
jiangzhengwenjz: GPT:
> I think `KirbyStartSwim` may still be a little too specific.
The function does not always enter `KirbySwim`. Its first branch is:
```c
if (kirby->base.base.base.unk62 & 4)
{
    sub_08058628(kirby);
    return;
}
```
> and `sub_08058628`/`sub_08058714` form a separate water-state pair (animation 56) which remains active while that collision condition holds. Horizontal movement transitions to the analogous animation-57 state (`sub_08058F3C`/`sub_08059008`), and both states call this function only after `unk62 & 4` clears or when Kirby pushes off into free movement.
Only the other path in this function actually installs `KirbySwim` as the update and sets animation 59.
This also matters at generic callers such as `Macro_0803FF64_6`, which invoke this function merely because `unk58 & 2` is set; they don't establish that Kirby is already in the free-swimming case.
So the description's claim that this function is simply the "swim-state entry" isn't quite true. I think it would be worth either using a slightly broader water-movement/state name for `sub_08059810`, or naming the adjacent 56/57 states first so we can describe the distinction confidently. `KirbySwim` itself looks well supported.

Basically you need to reconsider the function name. `sub_08058628` dispatches other movement pattern and can't be simply ignored.

jiangzhengwenjz (changes requested):
The PR is currently behind master as well

src/kirby.c
jackmgaller: Agreed on both counts — the dispatch through `sub_08058628` is real, and the old name described only one of the two paths. Reworked in the new revision (rebased onto current master):

- `sub_08059810` is now **`KirbyStartWaterMovement`**: the shared entry that routes Kirby into the correct in-water state — grounded (`unk62 & 4`) → the anim-56 state, otherwise anim 59 + `KirbySwim`.

  To make sure "water" is not over-claimed, I checked every caller rather than sampling: of 117 call sites in `kirby.c`, 113 sit directly under an `unk58 & 2` check (overwhelmingly the stereotyped `if (unk58 & 2) { KirbyEnterWater(kirby); KirbyStartWaterMovement(kirby); }` block), and the remaining 4 are inside the water idle/walk updates themselves, which only run while `unk58 & 2` holds. Both `dark_meta_knight.c` sites are gated the same way, and the one assembly caller in `cutscene_trigger.s` tests `[r2, #0x58] & 2` before the `bl`. So all 120 call sites (117 + 2 + 1) are provably in-water contexts.

- Following your second suggestion, the adjacent grounded states are now named too, using the `StartIdle`/`Idle`/`StartWalk`/`Walk` convention from the Bonkers states:
  - `sub_08058628` / `sub_08058714` → **`KirbyStartWaterIdle`** / **`KirbyWaterIdle`** — anim 56, standing on the bottom. The update hands off to the walk state on left/right input (`movementState & 0x20`/`0x10`) unless blocked by a wall (`unk62 & 1`), and pushes off into `KirbyStartWaterMovement` on jump input or when ground contact is lost.
  - `sub_08058F3C` / `sub_08059008` → **`KirbyStartWaterWalk`** / **`KirbyWaterWalk`** — anim 57. Entry and update both call `sub_0805B644`, which has no other callers, so it is these states' private movement routine; the update zeroes `xspeed` on wall contact. `sub_0805B644` itself is left unnamed in this PR.

If you'd prefer a different suffix than "Movement" (e.g. `KirbyStartWaterState`), happy to change it — the mechanism claim is the same either way.

jiangzhengwenjz (approved):

jiangzhengwenjz:
```
1424856 total bytes of code
```

jiangzhengwenjz:
```
41547 total symbols
    6792 symbols documented (16.3478%)
    34755 symbols undocumented (83.6522%)

14512864 total bytes of data

```

PR #386, opened 2026-08-09
Dissolve union Kirby_0 — resolves #100

### LINKED ISSUE #100: TODO: confirm that struct Kirby is not a union
I think `struct Kirby` doesn't contain a union. Basically, it has to be a union if `ObjectBase::kirby1` is really of type `struct Kirby *`, but actually it's likely capable of pointing to different types based on `struct Object2` (and,  it really makes no sense to make the field a union as it's accessed everywhere). 
```
src/bonus.c:298:    obj->base.kirby1 = arg0;
src/object.c:1179:        r4->base.kirby1 = r7->base.base.base.parent; // unknown type
src/object.c:1281:    r12->base.kirby1 = r4->base.base.base.parent; // unknown type
src/blockin.c:41:        arg0->base.kirby1 = sub_0803D368(&arg0->base); // It really returns `struct Kirby *` here. 
```
Fixing the type is easy so I will wait until more code is decompiled to get it more evident.

jackmgaller:
# Dissolve `union Kirby_0` — resolves #100

**Rebased onto master now that #385 has merged**, so the diff is this PR's own work
again: `Dissolve union Kirby_0` plus three commits answering your review. Six ROMs OK
from a wiped `build/`, `katam.gba` `274b102b6d940f46861a92b4e65f89a51815c12c`.

| commit | what |
|---|---|
| `Dissolve union Kirby_0` | the change itself |
| `Address review: type sub_080AC5E0's Kirby parameter as ObjectHeader *` | your two `src/soarar.c` notes — the tagged-pointer parameter, and `ObjType5ETo6C` in place of the open-coded range check |
| `Address review: read Mr. Frosty's ice cube through one typed local` | your `src/mr_frosty.c` note — the cast replaced by correcting `kirby3`'s type |
| `Document why sub_0809C48C's unk6C is safe to read as a Kirby` | your `src/object.c` provenance question, traced below |

You asked me to resolve #100 before doing more Kirby/object naming. Your 2022 hypothesis
was right: `struct Kirby` does not contain a union, and the union is an artifact of
mistyped pointers.

## Why the union is fake

Its two arms disagree about bytes **0x7C–0xB4** — `struct Object` puts
`type`/`kirbyAbility`/…/`objTemplate` there, `struct Kirby_0_1` puts two `struct Sprite`.
So the union claims Kirby uses those bytes two ways. It doesn't:

- **Genuine Kirbys never read that range through the `struct Object` arm.** Every
  `gKirbys[..]` access through it is `.base` (276 sites, all `ObjectBase`, below 0x78)
  or `.unk78` (3 sites, at exactly 0x78 — the other arm's `filler78`, which does not
  overlap the sprites).
- Offset 0x78 holds a **Kirby state function**. All 196 functions assigned to it are
  declared `(struct Kirby *)`, and the one non-function assignment is `gUnk_0834C4B0[]`,
  which the tree already declares `void (*const [])(struct Kirby *)`.

Result:

```c
struct Kirby {
    /* 0x000 */ struct ObjectBase base;
    /* 0x078 */ void (*stateFn)(struct Kirby *);
    /* 0x07C */ struct Sprite sprites[2];
    /* 0x0CC */ struct Task *task;
    ...
```

## The four reads that did need the Object arm

Each came through a pointer *declared* `struct Kirby *` that holds a `struct Object *` —
exactly the mistyping you predicted — and three are guarded by an explicit tag test, so
the code already documented its own polymorphism:

```c
// src/object.c, where r6 is `struct Kirby *r6 = r5->base.unk6C;`
} else if (r6->base.header.kind != 1) {   // not a struct Object -> bail
    return;
} else if (((struct Object *)r6)->type) {
```

`src/mr_frosty.c` short-circuits on `kind == 1` the same way.

**The fourth is weaker and I want to flag it rather than have you find it.** In
`src/soarar.c`, `sub_080AC5E0`'s `else` arm only rules out `kind == 0`. Tracing its only
caller — `sub_0803699C`'s kind-2 dispatch, reached from `CreateSnooter` →
`sub_080AC8CC`, which allocates `sizeof(struct ObjectBase)` — a `kind == 2` pointer to a
**0x78-byte** allocation can reach it, and `->type` is at 0x82. That is the original's
behaviour and is unchanged here; the union never protected it either. The comment there
now says only what the guard proves. `sub_0803699C` itself is more careful, testing
`kind == 1` before its own read at +0x82.

## Mistyped parameters fixed rather than cast around

22 call sites passed `&kirby->base.base` to a parameter declared `struct Object *`.
Rather than cast, I checked: **all 21 of `sub_08088640`'s call sites pass a Kirby**, it
only reads `->base.unk56`, and it stores the pointer into a `void *` that its consumer
reads back as `struct Kirby *`. So the parameter was simply wrong — retyping it removed
21 casts. `CreateObjTemplateAndObjWithSettingParent` needs none either: its macro only
uses `(obj2)->base.unk56`, which Kirby now has directly.

Four casts remain, all at the genuinely polymorphic sites above.

## Verification

- **Layout-neutral.** `sizeof(struct Kirby)` is `0x1A8` before and after, with `base` at
  0, `stateFn` at 0x78, `sprites` at 0x7C, `task` at 0xCC — checked by compiling
  `offsetof` probes against both trees. (A development check; there is no static-assert
  facility in the repo to land it in. Say the word if you'd like one added.)
- **All six ROMs OK from a wiped build**, `katam.gba` =
  `274b102b6d940f46861a92b4e65f89a51815c12c`, with a negative control.
- **The sweep is provably mechanical.** 80 of the 86 changed files reconstruct exactly
  from the parent by applying only five documented substitutions; the other six are
  hand-edited and small.
- **Assembly corroborates the layout independently of the ROM match**, which is worth
  more than either on its own: `asm/kirby_cannon.s` and `asm/code_08032E98.s` hard-code
  the `0x1A8` stride for `gKirbys[i]` (and `linker.ld` leaves `0x6A0` = 4 × 0x1A8);
  `asm/code_08032E98.s:347` loads `gKirbys+0x7C` and stores it as a sprite's
  `tilesVram`, i.e. the assembler reads 0x7C as `sprites[0]`, not `Object::unk7C`;
  `:10903` compares `gKirbys+0x78` against `sub_080566E0`, a `void (struct Kirby *)`.
  No asm site reads a Kirby at 0x80/0x82/0xB0.

**8,955 `base.base.base.` chains collapse to `base.`** — that triple was exclusively
Kirby, since the only other `struct Object base` member
(`include/destroyable_rock_block.h`) is two deep. That is the nesting you called
confusing.

## Two things I'd like your call on

**1. `stateFn` may be the wrong word.** It is a spelling this tree has not used. Every
other per-object state-function pointer is offset-named, including `struct Object::unk78`
— the same slot one struct over — and `ObjectSetFunc`, which sets exactly that field,
calls its parameter `func`. So `func` is arguably the established vocabulary and
`stateFn` a second one for it. I chose the semantic name because the field is no longer
unknown, but I will switch to `func` without argument, and `Object::unk78` probably wants
the same name in a follow-up.

**2. A follow-up I deliberately left out.** `ObjectBase::kirby2` (`include/data.h`) is
declared `struct Kirby *`, but assembly stores a `struct Object *` into it —
`asm/code_08032E98.s:9745`, `:10155`, `:10502` in `sub_08038010`, called from
`sub_08037314` with `gUnk_02022EC0` entries — and its C consumer reads only
`base.xspeed`/`base.yspeed`. It should be `void *` like its neighbours `unk6C` and
`parent`. Same defect class as the `kirby1` you listed in #100. I kept it out to keep
this reviewable; happy to do it next.

---

## `sub_0809C48C`'s `unk6C` — the provenance trace you asked for

You asked whether a kind-1 value is genuinely reachable, since `r6` is declared
`struct Kirby *` and read as a full Kirby well before the tag check. **It is not, and the
reason is structural rather than incidental.**

`sub_0809C380` is the only place that installs this callback, and it has exactly one
caller: `sub_080365C8`, `asm/code_08032E98.s`. That caller loads `r7->parent` into `ip`
and branches on its tag straight away:

```asm
_08036654:
	ldr r2, [r7, #0x70]   @ parent
	mov ip, r2
	...
_08036660:
	ldrb r0, [r1]         @ ip->kind
	cmp r0, #1
	bne _080366B4         @ kind != 1 --> ... --> str ip,[r4,#0x6c]; bl sub_0809C380
	...                   @ kind == 1 --> _0803668A --> str ip,[r5,#0x6c]; bl sub_0809CFC4
```

Both arms write `unk6C`, but they install *different* callbacks: the kind-1 arm goes
through `sub_0809CFC4`, whose last act is `unk78 = nullsub_123`. So:

| `parent->kind` | `unk6C` | `unk78` |
|---|---|---|
| 1 | the kind-1 pointer | `nullsub_123` |
| 0, 2, 3 | the parent | `sub_0809C48C` |

Reaching `sub_0809C48C` therefore means `unk6C` was never kind 1, and the `kind == 1` arm
inside it is the original's own dead code.

This is not a one-site coincidence — it is the idiom. **`unk6C` and `unk78` are written as
a pair**, so the callback identifies the tag. The other two `unk6C = ...->parent` writers
in `src/object.c` do the same, setting `sub_0809C994` and `sub_0809CE80` in the same
block.

**I tried your implied fix and it does not match.** Retyping `r6` to
`struct ObjectHeader *` with a Kirby view — the shape you asked for on `sub_080AC5E0` —
gives `katam.gba: FAILED`. It keeps a third pointer live alongside `r6`, the same cost
measured on `sub_0807A7E8` in #385. So the declaration stands and the invariant is a
comment instead. Say the word if you would rather have the cleaner shape and a
NONMATCHING.

**One gap, stated rather than hidden.** The install path excludes kind 1 but not kinds 2
and 3, and `animationIndex` (0xD4) and `inhaling` (0xDE) are read before any tag check —
both well past `sizeof(struct ObjectBase)` = 0x78, so either would be an over-read. That
the original reads them unconditionally is good evidence it only ever sees a Kirby here,
but I have not proven kind 0 from the disassembly, only kind != 1.

---

AI-assisted (Claude). Every claim above was checked against the code, and the change was
verified against the ROM.


src/soarar.c
```diff
@@ -527,7 +527,11 @@ bool32 sub_080AC5E0(struct Object2 *obj, struct Kirby *kirby) {
         }
     }
     else {
-        if ((u8)(kirby->base.base.type - 0x5E) > 0xE) {
+        // Not a Kirby -- the parameter is a tagged pointer. The guard only
+        // rules out kind 0, and sub_0803699C's kind-2 dispatch can reach here,
+        // so this read of ->type is the original's and is not proven safe by
+        // the tag alone.
+        if ((u8)(((struct Object *)kirby)->type - 0x5E) > 0xE) {
```
jiangzhengwenjz: You can probably use `ObjType5ETo6C` here

src/soarar.c
```diff
@@ -507,16 +507,16 @@ void sub_080AC45C(struct Object2 *obj) {
     }
 }
 
-bool32 sub_080AC5E0(struct Object2 *obj, struct Kirby *kirby) {
+bool32 sub_080AC5E0(struct Object *obj, struct Kirby *kirby) {
```
jiangzhengwenjz: The second parameter must be changed since kirby is a tagged pointer. `struct ObjectHeader *` should work

src/object.c
```diff
@@ -1182,19 +1182,19 @@ void sub_0809C380(struct Object2 *r3) {
...
     r3->base.flags &= ~0x20;
-    r3->unk86 = r4->base.base.base.unk56;
+    r3->unk86 = r4->base.unk56;
     r3->unk78 = sub_0809C48C;
     r3->base.unkC |= 0x400;
 }
 
-static void sub_0809C48C(struct Object2 *r5) {
+static void sub_0809C48C(struct Object *r5) {
     struct Kirby *r6 = r5->base.unk6C;
```
jiangzhengwenjz, line 1202: GPT:
> I think `sub_0809C48C` deserves one more pointer-provenance check.
The new `((struct Object *)r6)->type` cast itself looks appropriate because it is reached only after `kind == 1`. What concerns me more is that `r6` is globally declared as `struct Kirby *` and is dereferenced as a full Kirby (`animationIndex`, `inhaling`, etc.) well before that tag check.
If a kind-1 value is genuinely reachable here, those earlier Kirby accesses would mean the pointer model is still too narrow. If it is not reachable, then it would be useful to understand why the kind-1/Object branch exists and why the PR describes this as a genuinely needed Object-arm read.
Could you trace the possible values placed in `r5->base.unk6C` on this path before deciding whether `struct Kirby *r6` is still the right declaration?

src/mr_frosty.c
```diff
@@ -961,12 +961,12 @@ static void sub_080CD588(struct Task *t)
...
-static void sub_080CD594(struct Object2 *mf)
+static void sub_080CD594(struct Object *mf)
 {
-    if (mf->kirby3->base.base.base.unk0 == 1
-      && (mf->kirby3->base.base.type == OBJ_MR_FROSTY_ICE_CUBE_SMALL
-            || mf->kirby3->base.base.type == OBJ_MR_FROSTY_ICE_CUBE_LARGE)
-      && mf->kirby3->base.base.base.flags & 0x1000)
+    if (mf->kirby3->base.header.kind == 1
+      && (((struct Object *)mf->kirby3)->type == OBJ_MR_FROSTY_ICE_CUBE_SMALL
+            || ((struct Object *)mf->kirby3)->type == OBJ_MR_FROSTY_ICE_CUBE_LARGE)
```
jiangzhengwenjz: Instead of an explicit cast, the type of `kirby3` should be corrected.

jiangzhengwenjz (changes requested):

src/soarar.c
jackmgaller: Done in `804737f`. The parameter is `struct ObjectHeader *a2`, with the `Kirby` and `Object` views taken from it inside the function:

```c
bool32 sub_080AC5E0(struct Object *obj, struct ObjectHeader *a2) {
    struct Kirby *kirby = (struct Kirby *)a2;
    struct Object *obj2 = (struct Object *)a2;

    ...
    if (a2->kind == 0) {
```

and `obj->base.unk6C = a2;` at the end rather than storing the `Kirby` view. That also let the "not a Kirby" comment shrink, since the signature now says it.

src/soarar.c
jackmgaller: Adopted in the same commit — `if (!ObjType5ETo6C(obj2))` in place of the open-coded `(u8)(type - 0x5E) > 0xE`. Matches, and it is the same predicate spelled properly.

src/mr_frosty.c
jackmgaller: Done in `913230f` — one typed local instead of three casts:

```c
struct Object *ic = (struct Object *)mf->kirby3;
```

You are right that the cast was covering for the wrong type, but I could not fix it at the declaration the way you probably intended: `kirby3` is a `struct Kirby *` member shared with `gobbler.c` and `mega_titan.c`, which do use it for an actual Kirby. Correcting it there is a cross-file change to the member, not to this function, so I took the local and left a note saying what `kirby3` really holds on this path — every site that installs `sub_080CD594` as `unk7C` (`sub_080CD47C`, `sub_080CD4D0`, `sub_080CD524`) calls `sub_080CCEE0` first, and that is what stores the cube.

Happy to retype the member across all three files as a follow-up if you would rather have it fixed properly — it looks like the same defect class as the `kirby1` you listed in #100.

src/object.c
jackmgaller: Traced, and the answer is no — a kind-1 value is not reachable here. Written up in `662d9b7` and in the PR description; the short version:

`sub_0809C380` is the only installer of this callback, and it has exactly one caller, `sub_080365C8` in `asm/code_08032E98.s`. That caller loads `r7->parent` into `ip` and branches on the tag immediately:

```asm
_08036654:
	ldr r2, [r7, #0x70]   @ parent
	mov ip, r2
	...
_08036660:
	ldrb r0, [r1]         @ ip->kind
	cmp r0, #1
	bne _080366B4         @ kind != 1 --> ... --> str ip,[r4,#0x6c]; bl sub_0809C380
	...                   @ kind == 1 --> _0803668A --> str ip,[r5,#0x6c]; bl sub_0809CFC4
```

Both arms write `unk6C`, but they install different callbacks — the kind-1 arm goes through `sub_0809CFC4`, whose last act is `unk78 = nullsub_123`. So reaching `sub_0809C48C` means `unk6C` was never kind 1, and the `kind == 1` arm inside it is the original's own dead code.

It is not a one-site coincidence either: `unk6C` and `unk78` are written as a pair throughout, so the callback identifies the tag. The other two `unk6C = ...->parent` writers in `src/object.c` do the same, installing `sub_0809C994` and `sub_0809CE80` in the same block.

**I tried the fix your note implies and it does not match.** Retyping `r6` to `struct ObjectHeader *` with a `Kirby` view — the shape you asked for on `sub_080AC5E0` — gives `katam.gba: FAILED`: it keeps a third pointer live alongside `r6`, the same cost I measured on `sub_0807A7E8` in #385. So the declaration stands and the invariant is a comment instead. Say the word if you would rather have the cleaner shape and a NONMATCHING.

**One gap, which I would rather state than have you find.** The install path excludes kind 1 but not kinds 2 and 3, and `animationIndex` (0xD4) and `inhaling` (0xDE) are read before any tag check — both well past `sizeof(struct ObjectBase)` = 0x78, so either would be an over-read. That the original reads them unconditionally is good evidence it only ever sees a Kirby here, but I have proven `kind != 1` from the disassembly, not `kind == 0`.

src/kirby.c
jiangzhengwenjz, line 1: Github web interface refuses to show the file content so I have to make a file-level comment.
There are still `((void (*)(struct Kirby *))kirby->stateFn)` in the file where the cast is not necessary.

src/object.c
jiangzhengwenjz: GPT:
> You already called out the remaining gap here, so I don't think the current evidence is wrong. Before we leave this as an unresolved provenance TODO, could you do one last targeted check for whether kind == 0 can actually be established?
In particular, please trace all assignments/installers that can make unk78 == sub_0809C48C, including indirect/helper/assembly paths if any, and follow the corresponding unk6C value back to its source. The current path proves kind != 1, but I'm wondering whether an earlier invariant on the value stored in unk6C can rule out kinds 2 and 3 as well.
The unconditional reads of animationIndex and inhaling are certainly strong evidence that the original code expected a Kirby here, but I'd prefer to distinguish that source-intent evidence from an actual reachability proof.
If an exhaustive writer/installer trace still only establishes kind != 1, then I think keeping the Kirby view for matching and documenting the unresolved provenance with a TODO is the right outcome.

src/soarar.c
```diff
@@ -507,16 +507,22 @@ void sub_080AC45C(struct Object *obj) {
     }
 }
 
-bool32 sub_080AC5E0(struct Object *obj, struct Kirby *kirby) {
+// a2 is a tagged pointer: a struct Kirby when kind is 0, and something with a
+// struct Object prefix otherwise. kirby and obj2 are the two views of it; only
+// the one the kind check selects may be read past the header.
+bool32 sub_080AC5E0(struct Object *obj, struct ObjectHeader *a2) {
+    struct Kirby *kirby = (struct Kirby *)a2;
+    struct Object *obj2 = (struct Object *)a2;
```
jiangzhengwenjz: I asked GPT to do a static analysis for reachability:
> I traced the kind-2 dispatcher further, and the over-read can actually be made concrete.
The 0x78-byte helper created by sub_080AC8CC is the second operand of sub_0803699C, so it is not itself the value passed as a2. However, sub_08037314 performs bidirectional kind-2 × kind-2 dispatch through gUnk_0834BD88, whose kind-2 entry is sub_0803699C.
A concrete first operand is the Bomb object created by sub_08082380: it is allocated as exactly sizeof(struct ObjectBase) (0x78), has kind == 2, retains 0x40000000 for the normal Bomb-throw variants 0/1/2, and is registered in the kind-2 list through sub_0806F8BC -> SetPointerSomething.
If that Bomb object overlaps the Snooter helper, the reverse-direction dispatch is sub_0803699C(bomb, snooterHelper). The helper's parent is the Snooter, and the remaining masks also pass with their initialized values (helper->unk68 == 0x23, bomb->unk5C == 0, and the Bomb does not have 0x8000 set), so the type-7/0xE branch calls sub_080AC5E0(snooter, bomb).
a2->kind is then 2, so ObjType5ETo6C((struct Object *)a2) reads Object::type at +0x82 from an allocation that ends at +0x78. I think this is sufficient to document as a reachable original out-of-bounds read/UB, rather than merely saying that the tag does not prove the access safe.
Please also adjust the leading comment: kind != 0 still does not imply a struct Object prefix.

jiangzhengwenjz (changes requested):

src/kirby.c
jackmgaller: Fixed in `748cf9c` — both are gone, `kirby->stateFn(kirby)` now.

You are right that they were pure noise: `stateFn` has been declared `void (*)(struct Kirby *)` since this branch dissolved the union, so the cast named the pointer's own type. They were left over from when the field was reached through the `Object` arm and genuinely needed casting. Six ROMs OK.

(Apologies for the file-level workaround being necessary — `src/kirby.c` is one of the big ones in this diff.)

src/soarar.c
jackmgaller: Adopted, `8faa77d`, both halves.

I checked the chain rather than taking it as read, and it holds:

- `struct Object::type` is at **+0x82**; `sizeof(struct ObjectBase)` is **0x78**. So the read is 11 bytes past the end.
- `sub_080AC5E0`'s only caller is inside `sub_0803699C`, and at that site (`asm/code_08032E98.s`, `_08036BA2`) the second argument is `r4` — `sub_0803699C`'s own first operand — exactly as your trace says.
- The bomb from `sub_08082380` (`src/code_0806F780.c:7361`) is `TaskCreate(..., sizeof(struct ObjectBase), ...)` with `header.kind = 2`, and it does keep `0x40000000` (set as part of `flags |= 0x60000040`).

So this is a reachable out-of-bounds read in the original, not merely an unproven one, and it now says so. The leading comment no longer claims kind != 0 implies a `struct Object` prefix — it says any non-zero kind guarantees only the header, and that kind 2 is frequently a bare `struct ObjectBase`.

Transcribed as it stands rather than guarded, since the ROM does it. Say the word if you would rather it were guarded and NONMATCHING.

src/object.c
jackmgaller: Done — and the trace goes further than `kind != 1`. **It reaches `kind == 0`.** `ce37eed`.

First, a correction to my own last comment, since it is what your question was probing. I wrote that "its writers always set `unk78` alongside it". That is **not literally true**: `_08036720` and `_08036758` in `sub_080365C8` both store `unk6C` and leave `unk78` untouched. Same arm and same provenance, so the conclusion survived, but the sentence overclaimed and you would have found it.

Redone properly, in the order you asked for.

**The installer set is closed, at the binary level.** `sub_0809C48C` is `static` and assigned exactly once in the tree. Stronger: **`0x0809C48D` occurs exactly once in the whole 16MB ROM**, at file offset `0x9C488` — the literal pool immediately before `sub_0809C48C`, i.e. the constant `sub_0809C380` loads. There is no second installer, no table entry and no copy, in decompiled code or in anything still `.incbin`. That answers the "indirect/assembly paths if any" half outright.

**The only caller is a table slot.** `sub_0809C380` has one `bl`. Its caller `sub_080365C8` has *no* direct `bl` anywhere — `0x080365C9` also occurs exactly once in the ROM, at `0x34BD8C`, which is `gUnk_0834BD88[1]`. The table decodes as:

```
[0] 0x080364E5   [1] 0x080365C9 (sub_080365C8)
[2] 0x0803699D   [3] 0x0812424D
```

and `sub_08037314` calls it as `table[self->kind](other, self)` — indexed by the **second** argument's kind. So in `sub_080365C8`, `r5` is the kind-1 object and `r7` is the other participant.

**The gate is what closes it.** The install path is reachable only when `r7->unk68 & 0x20` (`_08036646`). I swept every writer of that bit — 9 in `src/`, 3 in `asm/` (`leap.s`, `cookin.s`, `prank.s`) — and **every one is a kind-2 inhale zone whose `parent` is assigned once at creation and never rewritten**. Those parents are Kirbys (`sub_08071FC0`, `sub_08072314`) or `struct Object`s (`sub_080C89E0`, `sub_080D4FCC`, `sub_080CD050`, `sub_080CF20C`, `sub_080A09A4`, `sub_080E761C`, `sub_080AC8CC`, plus the three asm ones via `InitObject`). Nothing else.

And the `struct Object` half is precisely what `_08036660` diverts to `sub_0809CFC4`. What is left on the `sub_0809C380` arm is the Kirby half.

Kind 3 is structurally impossible here for a separate reason worth recording: registration goes through `gUnk_02022EB0[room][kind - 1]` with `u8 gUnk_02022EB0[][2]` (`include/data.h:645`, `src/code_0806F780.c:751`), so the interaction lists hold **only kinds 1 and 2**. A kind-3 `r7` would in any case have over-read at `unk68` (+0x68) before ever reaching `parent`, since `sizeof(struct EffectObject)` is 0x48.

So the unconditional `animationIndex` / `inhaling` reads are sound, and `struct Kirby *r6` is the right declaration.

**Two caveats I would rather state than have you find:**

1. **This is a whole-program sweep, not a local invariant.** It holds only while that `unk68` bit-0x20 writer set stays complete. If a future decomp introduces a kind-2 zone whose parent is kind 2, the argument breaks silently — which is why it is written into the comment rather than left implicit. If you would still prefer a `TODO` marker on top of that, I am happy to add one; I left it off only because the trace did not bottom out where you expected it might.
2. **Post-install clobbering is proof by survey, not by construction.** `unk78` is a deferred callback, and I did not exhaustively prove nothing rewrites `r5->unk6C` between install and invocation. Every `unk6C` writer I found stores a Kirby, `NULL`, or a same-provenance `->parent`, so nothing contradicts it, but that is weaker than the rest of the argument.

jiangzhengwenjz (approved):

jiangzhengwenjz:
```
1424856 total bytes of code
```

jiangzhengwenjz:
```
41547 total symbols
    6792 symbols documented (16.3478%)
    34755 symbols undocumented (83.6522%)

14512864 total bytes of data

```

PR #366, opened 2026-07-20
Decompile chest, snooter, soarar, star_platform, special_hub_mirror, large_stone_block_7D

jackmgaller:
Decompiles 6 asm files (54 functions) to C, as discussed on Discord:

| File | Functions | Status |
|---|---|---|
| src/soarar.c | 21 | all matching |
| src/chest.c | 10 | 9 matching, sub_0800AEB0 NONMATCHING |
| src/snooter.c | 7 | all matching |
| src/special_hub_mirror.c | 7 | all matching |
| src/star_platform.c | 6 | all matching |
| src/large_stone_block_7D.c | 3 | all matching |

`make compare` passes (`katam.gba: OK`).

Notes:
- sub_0800AEB0 uses the existing NONMATCHING convention (naked function including `asm/nonmatching/sub_0800AEB0.inc`, functionally equivalent C behind the ifdef, same as flamer.c) — its register allocation resisted everything I tried.
- snooter.s and soarar.s cross-reference each other's functions and were likely one source file originally (as the old asm comment suggested); they're kept as separate files matching the existing split.
- No renames beyond what decompilation requires; new functions keep sub_/gUnk_ names.

As disclosed on Discord, this work was AI-assisted (Claude), with every function verified byte-matching against baserom via the build.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #387, opened 2026-08-09
Decompile cutscene_trigger.c, 91 functions / 6525 asm lines

jackmgaller:
Decompiles all 91 functions from `asm/cutscene_trigger.s` into `src/cutscene_trigger.c`, replacing its linker entry with `src/cutscene_trigger.o`. The original cutscene behavior and ROM layout are preserved. This work was AI-assisted; the latest review fixes were implemented with Codex and independently checked by a subagent.

Current master (`52fc58f`) is merged without rewriting the PR history. The cutscene code now uses the current Kirby layout and the existing five-argument `Macro_081050E8`.

The review fixes simplify all four movement clamps, use `bool32` for `allDone`, replace the staged speed clamp with `min(trigger->unkB4 / 2, 0xC8)`, remove the three redundant table pointers and the temporary flags variable, and use separate compound flag assignments throughout analogous sites. Palette setup uses the existing macro, preserving each variant and condition. Ability, save-buffer, and blend-register values use the established constants. Indentation was checked across the complete file.

The two animation tables at `0x082DE9FC–0x082DEA24` and `0x082DEA24–0x082DEA4C` are declared as ten `CutsceneAnimPair` records, each containing two `u16` fields at offsets 0 and 2. Their consumers use the first field as the animation ID and the second as the variant. The tables at `0x082DEA6C`, `0x082DEA7C`, and `0x082DEA8C` each occupy 0x10 bytes and are declared `s16[4][2]`; `0x082DEA9C` occupies 8 bytes and is `s16[4]`. The interleaved x/y ranges `0x082DEA4C–0x082DEA5C` and `0x082DEA5C–0x082DEA6C` retain their original bytes while their split assembly labels are consolidated. Subtype 5's `unkCC` is a Boolean flag for pending camera-offset restoration, so it now uses `bool8` and `TRUE/FALSE`; subtype 0's separate bitmask remains unchanged.

At the reviewer's request, seven additional functions referenced by `gUnk_08351648` are exported and declared in their owning headers: `BonusSetFunc`, `sub_08123780`, `WaddleDee37ChooseXSpeed`, `sub_080AB720`, `sub_080B6A54`, `sub_080C0CBC`, and `sub_080C2B28`. These cover 18 table entries. The final ELF audit confirms that all 217 non-null handlers in the 219-record table resolve to global Thumb functions at their original addresses.

`CreateCutsceneTrigger` deliberately retains the original uninitialized `size/dtor` path for subtype 3 or values above 6. A local TODO records that known spawners use 2, 4, 5, or 6, while other runtime templates have not been ruled out. The previous claim that the invalid path was proven unreachable was too strong. Unknown symbols and unreviewed source-form constraints remain conservative; the removed clamp temporaries and byte staging are no longer claimed to be necessary.

Validation on final commit `233e6e8`:
- Clean, dependency-aware `make compare -j8` after `make tidy`, using the bundled agbcc and GNU coreutils on macOS: all six checks pass (`katam.gba`, `payload.gba`, `speed_eaters.gba`, `unk_8D94B9C.gba`, `unk_8E1FE28.gba`, `unk_8E8490C.gba`). Neither `NODEP=1` nor `MODERN=1` was used.
- Full clean build with `-DNONMATCHING` added to the existing preprocessor flags: compiles and links. Temporary build-file changes were restored before the final matching build.
- Instruction and relocation comparison with `objdump -dr` for the cutscene object after the individual simplifications; no differences from the matching merged baseline.
- `git diff origin/master --check` passes; working tree is clean.
- Independent source/assembly review found no functional regression in the critical changes. This did not include emulator gameplay testing.


src/cutscene_trigger.c
```diff
@@ -0,0 +1,1939 @@
...
+        trigger->unkC4->unkA = 0x80;
+        trigger->unkC4->unk4 = 0xFFFF;
+        trigger->unkC4->unk6 = -1;
+        trigger->unkB4 = 0;
+        trigger->obj2.unk78 = sub_080233A8;
+        sub_0808859C(&trigger->obj2, gUnk_0835105C[10], 0x48);
+    }
+}
+
+static void sub_08022DE4(struct Object2 *obj)
```
jiangzhengwenjz: The function pointer is referenced in one of the baserom incbins so it's not static. Check other functions as well and do not guess.

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1939 @@
...
+static void sub_080237F0(struct CutsceneTrigger1 *);
+
+void *CreateCutsceneTrigger(struct Object *template, u8 playerId)
+{
+    struct Task *t;
+    struct Object2 *obj;
+    u32 size;
+    TaskDestructor dtor;
+
+    switch (template->subtype1)
```
jiangzhengwenjz, line 114: I think this is worth further inspection. Is it possible that subtype1 is 3 or other value? Uninitialized use if so.

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1939 @@
...
+            bool32 done = TRUE;
+            s32 diff;
+            s32 spd;
+
+            if (o->x > kirby->base.base.base.x)
+            {
+                    diff = o->x - kirby->base.base.base.x;
+                    spd = (diff >> 8) * gUnk_082DEA6C[j][0];
+                    if (spd < gUnk_082DEA7C[j][0])
+                        spd = (u16)gUnk_082DEA7C[j][0]; // (u16) loads ldrh not ldrsh; required for matching (all four clamps)
```
jiangzhengwenjz: Let's try to remove the cast. Maybe `o->unk3C = spd;` is the result of a tail merge? Try other mutations too if it doesn't work. Experiment simultaneously with other similar parts of the function as well.

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1939 @@
...
+                    else
+                    {
+                        s32 max = gUnk_082DEA8C[j][0];
+                        s32 tmp = spd;
+
+                        if (tmp > max)
+                            tmp = max;
+                        spd = tmp;
```
jiangzhengwenjz: Super inefficient. Try to improve the code with maybe a variant of the current `max` macro

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1939 @@
...
+        obj->unk78 = sub_0802172C;
+        break;
+    case 4:
+        obj->unk78 = sub_08022F24;
+        break;
+    case 5:
+        obj->unk78 = sub_08022174;
+        break;
+    case 6:
+        {
```
jiangzhengwenjz: Are the brackets necessary? The indentation doesn't make sense even if so.

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1939 @@
...
+    bool32 done = TRUE;
+    u16 i;
+
+    for (i = 0; i < 4; i++)
+    {
+        if (trigger->unk294[i] != NULL)
+        {
+            struct Object4 *o = trigger->unk294[i];
+
+            if (o->x <= 0x11FFF && o->y > -0x2000)
```
jiangzhengwenjz: How about `< 0x12000` so that the expressions are more consistent

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1939 @@
...
+        o->unk3C = 0;
+        o->unk3E = 0;
+        o->sprite.unk8 = (o->sprite.unk8 & ~0x3000) | 0x1000;
+        n++;
+    }
+    PlaySfx(&trigger->obj2.base, SE_DARK_MIND_WARPSTAR_ENTER);
+    trigger->obj2.unk78 = sub_08021984;
+}
+
+static void sub_08021984(struct CutsceneTrigger2 *trigger)
```
jiangzhengwenjz, line 663: The indentation of the function is a mess

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1939 @@
...
+
+static void sub_08022D78(struct CutsceneTrigger5 *trigger)
+{
+    if (++trigger->unkB4 > 0x3B)
+    {
+        trigger->unkC4 = sub_0803C83C(5, trigger->obj2.base.roomId);
+        trigger->unkC4->unk8 |= 0x40;
+        trigger->unkC4->unkA = 0x80;
+        trigger->unkC4->unk4 = 0xFFFF;
+        trigger->unkC4->unk6 = -1;
```
jiangzhengwenjz: Is there any reason why 0xFFFF and -1 are used respectively for unk4 and unk6?

jiangzhengwenjz (changes requested):
I anticipate significant changes to this PR after #385 and #386 are merged, so for now, I have only performed a cursory review. A detailed review will follow later.

src/cutscene_trigger.c
jackmgaller: Fixed — `sub_08022DE4` is no longer `static`, and it is declared in `include/cutscene_trigger.h` alongside `CreateCutsceneTrigger`, matching how `include/blipper.h` and `include/waddle_dee.h` declare theirs.

It is `gUnk_08351648[0x8E].unk10` — stride 0x18, field +0x10, declared `void (*)(struct Object *)`, which is exactly its signature.

**You said not to guess, so here is the audit rather than a spot fix.** For all 90 static functions in the file I took the address from the original `asm/cutscene_trigger.s` labels, searched the entire 16MB ROM for `addr | 1`, and discarded hits inside the file's own `.text` (`0x080206A0`–`0x08023810`), which are its literal pools.

Two came back, and only one is real:

| function | hits outside .text | verdict |
|---|---|---|
| `sub_080214E4` | `0xD8CDE7`, `0xDAF983` | **not references** — both unaligned, so they are byte coincidences inside the `0x08D8xxxx` data blobs |
| `sub_08022DE4` | `0x083523A8` | **real** — 4-byte aligned, and its neighbours in the table are sibling handlers (`0x08025321`, `0x08025435`) |

So exactly one function needed it, and the other 89 are genuinely file-local.

**One thing I want to put to you rather than decide myself.** I checked the whole table against the linked ELF: of its 217 handlers, **198 are global with a header prototype and 19 are `static`** — `BonusSetFunc`, `WaddleDee37ChooseXSpeed`, `sub_080AB720`, `sub_080B6A54`, `sub_080C0CBC` and others. I had also recorded from your review on #366 that a function referenced only from the object type table does not need exporting, since the table is raw data with no relocations.

I have followed what you asked for here, since it is also what 198 of the 217 do. But that leaves the 19 inconsistent. Do you want them swept to match in a follow-up, or is `static` acceptable there and this file is the exception?

src/cutscene_trigger.c
jackmgaller: Good instinct — **the hazard is real, but it is not reachable.** No code change; the C is a faithful transcription of it.

**The original has the same hole.** `CreateCutsceneTrigger @ 0x080206A0`:

```asm
	ldrb r0, [r6, #0xe]      @ subtype1
	cmp  r0, #6
	bhi  _080206FC           @ > 6 -- straight to the TaskCreate setup
	lsls r0, r0, #2
	...jump table...
	.4byte _080206FC         @ case 3 lands on the same label
```

`_080206FC` is the call sequence, and on that path neither `r5` (size) nor `r3` (dtor) has been written — they carry whatever the caller left. So `case 3` and `> 6` really do reach `TaskCreate` with both uninitialised, exactly as the C says. The missing `case 3` is the ROM's, not mine.

Downstream that is not benign: `TaskCreate` stores the pointer verbatim (`src/task.c:84`) and `CpuFill16` zeroes `size` bytes, then `TaskDestroy` does `if (task->dtor != NULL) task->dtor(task);` (`src/task.c:137`) — a wild indirect call. At the three call sites, `r5` is `playerId` in `CreateObject` and the template pointer in the other two, so `size` would be 0–3 or the low half of a ROM address.

**Reachability — it does not happen.** `CreateCutsceneTrigger` is `gSpawnFuncTable1[0x8E]` (`data/object.s:152`, matching `OBJ_CUTSCENE_TRIGGER`). Sources of a template for that type:

- **Room lists.** I walked all 287 `gLevelObjLists` entries: 3482 records, every one with `unk1 == 0x24` and `spawnTable == 1` — which is also my check that the walk did not desync. Exactly **two** have `type == 0x8E`: room 0x321 at `0x088B4AD8` with `subtype1 = 6`, and room 0x323 at `0x088B4DA8` with `subtype1 = 4`.
- **Runtime templates.** Two spawn sites, both literal: `src/dark_mind.c:6264` passes **2**, `src/dark_meta_knight.c:2799` passes **5**.
- **Undecompiled asm.** The six `movs rX, #0x8e` immediates left in `asm/` are all in `asm/code_0802B4A8.s` and are struct-offset computations, not object types.

So the reachable domain is **{2, 4, 5, 6}** — a strict subset of the handled set. Subtypes 0 and 1 are handled by both dispatchers but have no spawner at all; they look like dead cutscene variants.

Summary: the switch is exhaustive over everything that actually occurs, but nothing in the code enforces it. The limit of the proof is that the grep would miss a spawn that computes `0x8E` into a variable rather than writing it literally; I found no such site, but I cannot rule one out. Happy to add a short comment recording the {2,4,5,6} domain if you want it written down.

src/cutscene_trigger.c
jackmgaller: Removed, in all four clamps — and **your tail-merge hypothesis is what found it.**

Chasing it directly showed the narrowing really does come from where you suspected. Storing `o->unk3C` separately in each arm did reproduce the correct `ldrh` with no cast (598 instructions vs the target's 608), which confirmed the mechanism, but it is not the shape the ROM has: it emits `negs r0, r1` off the live register, whereas the per-arm store reloads with `ldrh` + `negs`. So the value really is held in a variable and stored once after the if/else — but the *floor arm alone* carries an unsigned-16 conversion.

That is what the fix expresses. Instead of a cast, a local whose declared type does the narrowing:

```c
if (spd < gUnk_082DEA7C[j][0])
{
    u16 lo = gUnk_082DEA7C[j][0];

    spd = lo;
}
```

Exact match, 608 instructions, and the apology comment is gone.

For completeness, the shapes that did not work, with instruction counts against the target's 608:

| shape | count |
|---|---|
| cast simply deleted | 600 |
| `s16` join variable + `min` | 616 |
| `u16` join variable + `min` | 612 |
| tail merge, store in each arm | 598 |
| tail merge + original ceiling block | 600 |
| `gUnk_082DEA7C[j][0] & 0xFFFF` | 624 |
| **`u16 lo` local** | **608 ✅** |

A `u16` *join* variable cannot work because the ceiling arm passes an untruncated 32-bit value into the store, so the joined value has to be `int`; only the floor arm is unsigned-16. Hence the narrow local rather than a narrower `spd`.

While I was here: there is one more comment of this kind in the file, at `sub_08022730` — `spd = trigger->unkB4 / 2; // the u8 t staging and clamp arm order are required for matching`. You did not flag it, but it is the same apology-in-a-comment shape, and this one worked out, so say the word and I will attack it the same way.

src/cutscene_trigger.c
jackmgaller: Tried this eight ways and **it does not match** — reporting it as a measured negative rather than leaving it silently unchanged.

What did ship: the local was named `max`, shadowing `max()` from `include/global.h`, which was genuinely confusing. It is now `hi`. That rename is free — 608 instructions, zero diff.

The structure has to stay. The two ceiling blocks in the x-decreasing and y-increasing directions compile to a register shuffle:

```asm
adds r0, r1, #0
cmp  ...
ble  ...
adds r0, r2, #0
adds r1, r0, #0
```

which only the two-local form reproduces. Everything leaner collapses it to the tight in-place form and loses 4 instructions — 604 against the target's 608, every time:

- `else if (spd > gUnk_082DEA8C[j][k]) spd = ...;` — the shape the *other* two blocks legitimately use
- `spd = min(spd, gUnk_082DEA8C[j][k]);`
- `spd = min(hi, spd);` and `spd = min(spd, hi);` with an `s32 hi`
- `s32 tmp = spd; spd = min(tmp, hi);`, and `tmp = min(tmp, hi); spd = tmp;`
- `spd = spd > hi ? hi : spd;`
- `min()` retried after the `u16 lo` change landed, in case register allocation had shifted — still 604

Both locals are load-bearing: inline the table access, or drop the `tmp` copy, and it is 604 either way.

So the block is inefficient, but the inefficiency is the 2004 compiler's, not the transcription's — and the asymmetry between the four blocks is real rather than sloppiness, since the two clamp blocks that *can* use the tight form already do. I did not add a new macro to `global.h` for it. If you would rather have the clean `min()` and a NONMATCHING, that is your call and I will make the change.

src/cutscene_trigger.c
jackmgaller: Removed. The block declared nothing, so the braces were doing no work, and as you say the brace column was wrong regardless. `case 6` now reads like its five siblings.

src/cutscene_trigger.c
jackmgaller: Changed to `o->x < 0x12000`. Same test, and it matches how the surrounding bounds are written. Still matching.

src/cutscene_trigger.c
jackmgaller: Reindented. It was: the bodies of the four `if (o->x/y ≷ kirby...)` arms carried a spurious extra level, `case 0:` opened a block whose brace sat at the wrong column, and `case 2`/`3`/`4` were inconsistent with each other.

Case labels now sit at the `switch`'s own indentation throughout, matching the `switch` at line 1170 of this file. The `(u16)` casts and the `max` shadowing you flagged separately are fixed in the same commit, so the whole function should read cleanly now.

I also swept the rest of the file for the same defect rather than waiting for you to find it: `sub_08022770` had seven more lines of it — control-statement bodies at or left of their header, plus a continuation line of a multi-line `if` aligned as though it were a statement. Fixed in a separate commit. Those two functions were the only ones affected; the file is clean now by that check.

src/cutscene_trigger.c
jackmgaller: No good reason — changed to `0xFFFF` for both.

Both fields are `u16` (`struct Unk_02022930_0`, `include/data.h:624-625`), so `-1` was an implicit conversion storing the identical halfword. It was the only `-1` among the writers of these fields; `src/dark_mind.c` and `src/dark_meta_knight.c` all use hex.

One thing that makes this less obvious than it looks, since it nearly made me leave it alone: the two stores really *are* encoded differently in the ROM — a literal-pool load for `+0x4`, a `movs`/`rsbs` pair for `+0x6`. But that is positional rather than caused by the spelling. `sub_0810C62C` (`src/dark_mind.c:6233-6234`) writes `0xFFFF` to both fields and compiles to the identical instruction pair. Verified by building: six ROMs OK, `katam.gba` unchanged.

src/cutscene_trigger.c
jiangzhengwenjz: Please change them as well, thanks.

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+                            {
+                                trigger->unkD2++;
+                                trigger->unkD4 = 0;
+                            }
+                            else
+                            {
+                                trigger->unkD0 = 0;
+                                trigger->unkD4 = 4;
+                            }
+                    }
```
jiangzhengwenjz, line 283: Obviously indentation is not fixed in the function and there are a lot of other cases. Please rescan the whole file and fix everything. Make sure you DO fix them.

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+            kirby->base.base.base.x += trigger->unkCE[i][0];
+            kirby->base.base.base.y += trigger->unkCE[i][1];
+            if (kirby->base.base.base.x >> 8 >= 0x168 && kirby->base.base.base.x >> 8 <= 0x178
+             && kirby->base.base.base.y >> 8 > 0xD7 && kirby->base.base.base.y >> 8 <= 0xE8)
+            {
+            kirby->base.base.base.x = 0x17000;
+            kirby->base.base.base.y = 0xE000;
+            kirby->base.base.base.flags |= 0x800;
+            trigger->unkDE[i] = 1;
+            }
```
jiangzhengwenjz: Again indentation

src/cutscene_trigger.c
jiangzhengwenjz: “it is not reachable” is stronger than what has actually been established. Please leave a short local TODO comment beside this switch recording the unresolved condition.

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+    u16 count = 0;
+    u16 i;
+
+    for (i = 0; i < 4; i++)
+    {
+        struct Kirby *kirby = &gKirbys[i];
+
+        if (kirby->hp <= 0)
+            continue;
+        if (trigger2->obj2.base.roomId != kirby->base.base.base.roomId)
```
jiangzhengwenjz: Outdated. Please merge the current `master`

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+    {
+        if (!(trigger2->unkCC >> i & 1))
+            continue;
+        gKirbys[i].base.base.base.flags &= ~0x1000000;
+        kirby = &gKirbys[i];
+        if (kirby->ability == 0xB)
+            sub_080641FC(kirby);
+        else if (kirby->ability == 0xE)
+            sub_0806A798(kirby);
+        else if (kirby->ability == 0x13 && kirby->base.base.base.flags & 0x40)
```
jiangzhengwenjz: use `KIRBY_ABILITY_*`

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+        o->y = sword->base.y - 0x2800;
+        comp->base.flags |= 0x1000;
+        trigger->obj2.unk78 = sub_08022EF8;
+    }
+}
+
+static void sub_0802172C(struct CutsceneTrigger2 *trigger)
+{
+    struct CutsceneTrigger2 *trigger2 = trigger;
+    const s16 (*xyTable)[2];
```
jiangzhengwenjz: We don't need the variable

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+        n++;
+        trigger2->unk2B4++;
+    }
+    trigger->obj2.unk78 = sub_08022F0C;
+}
+
+static void sub_08021844(struct CutsceneTrigger2 *trigger)
+{
+    struct CutsceneTrigger2 *trigger2 = trigger;
+    const s16 (*xyTable2)[2];
```
jiangzhengwenjz: We don't need the variable

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+    trigger->unkB8->x += trigger->unkC4;
+    trigger->unkB8->y += trigger->unkC6;
+    if (trigger->unkB8->y <= -0x4000)
+        trigger->obj2.unk78 = sub_08023054;
+}
+
+static void sub_08022174(struct CutsceneTrigger5 *trigger)
+{
+    struct CutsceneTrigger5 *trigger2 = trigger;
+    const u16 (*animTable)[2];
```
jiangzhengwenjz: We don't need the variable

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+
+    if (aiKirbyState < AI_KIRBY_STATE_TUTORIAL)
+    {
+        gAIKirbyState = AI_KIRBY_STATE_TUTORIAL;
+        if (!(gUnk_0203AD10 & 0x10))
+        {
+            if (gUnk_0203AD10 & 2)
+            {
+                if (gLocalPlayerId == gUnk_0203AD24)
+                    UpdateSaveBufferByOffset(1, gSaveID <= 2 ? gSaveID : 0);
```
jiangzhengwenjz: use `SAVE_BUFFER_TYPE_*`

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+                else
+                    sub_08031CE4(8);
+            }
+            else
+                UpdateSaveBufferByOffset(1, gSaveID <= 2 ? gSaveID : 0);
```
jiangzhengwenjz: use `SAVE_BUFFER_TYPE_*`

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+    {
+        animTable = gUnk_082DE9FC;
+        trigger->unkDE[i] = trigger->unkE2[i] = 0;
+    }
+    trigger2->unkB4 = 0;
+    animId = animTable[4][0];
+    variant = animTable[4][1];
+    trigger2->unkBC = sub_0808B62C(&trigger->obj2.base, 0x14, animId, variant, 0);
+    trigger2->unkBC->sprite.palId = 0;
+    if (gKirbys[gLocalPlayerId].base.base.base.roomId == trigger2->unkBC->roomId)
```
jiangzhengwenjz: Can you add a `variant` parameter to `Macro_081050E8` and use it here?

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+    else
+        trigger2->unkBC->sprite.palId = 0;
+    trigger2->unkBC->x = 0xB800;
+    trigger2->unkBC->y = 0x11C00;
+    trigger2->unkBC->flags ^= 1;
+    animId2 = gUnk_082DEA24[4][0];
+    variant2 = gUnk_082DEA24[4][1];
+    trigger2->unkC0 = sub_0808B62C(&trigger->obj2.base, 0x14, animId2, variant2, 0);
+    trigger2->unkC0->sprite.palId = 0;
+    if (gKirbys[gLocalPlayerId].base.base.base.roomId == trigger2->unkC0->roomId)
```
jiangzhengwenjz: same as above

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+    {
+        struct Kirby *kirby = &gKirbys[i];
+        u32 flags;
+
+        if (kirby->hp <= 0)
+            continue;
+        if (kirby->base.base.base.roomId != trigger->obj2.base.roomId)
+            continue;
+        flags = kirby->base.base.base.flags | 0x1000800;
+        kirby->base.base.base.flags = flags | 0x100;
```
jiangzhengwenjz: Are you sure we need the temp variable? Maybe it's just two statements with `|=`

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+        }
+        case 1:
+            if (++kirby->animationIndex > 0x59)
+                kirby->animationIndex = 0x4A;
+            if (trigger->unkE2[i] & 1)
+                kirby->base.base.base.flags ^= 0x400;
+            if (++trigger->unkE2[i] > 0x3B)
+            {
+            kirby->base.base.base.flags |= 0x400;
+            trigger->unkDE[i] = 2;
```
jiangzhengwenjz: indentation

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+            trigger->unkCE[i][1] = (a[0] * gSineTable[angle] + a[1] * gSineTable[angle + 0x100]) >> 14;
+            kirby->base.base.base.x += trigger->unkCE[i][0];
+            kirby->base.base.base.y += trigger->unkCE[i][1];
+            if (kirby->base.base.base.x >> 8 >= 0x168 && kirby->base.base.base.x >> 8 <= 0x178
+             && kirby->base.base.base.y >> 8 > 0xD7 && kirby->base.base.base.y >> 8 <= 0xE8)
+            {
+            kirby->base.base.base.x = 0x17000;
+            kirby->base.base.base.y = 0xE000;
+            kirby->base.base.base.flags |= 0x800;
+            trigger->unkDE[i] = 1;
```
jiangzhengwenjz: indentation

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+    struct EffectObject *o;
+
+    trigger->unkB4 = 0;
+    sprite = &trigger->unkBC->sprite;
+    sprite->animId = gUnk_082DE9FC[7][0];
+    sprite->variant = gUnk_082DE9FC[7][1];
+    sprite->unk1C = 0x20;
+    sub_08155128(sprite);
+    o = trigger->unkBC;
+    o->flags = (o->flags & ~4) | 8;
```
jiangzhengwenjz:
```suggestion
    o->flags &= ~4;
    o->flags |= 8;
```
How about this? There are other similar cases and please check all of them

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+    o->unk4 = 1;
+    o->header.unk2 = 0x20;
+    o->header.unk1 = 2;
+    sprite = &trigger->unkC0->sprite;
+    sprite->animId = gUnk_082DEA24[7][0];
+    sprite->variant = gUnk_082DEA24[7][1];
+    sub_08155128(sprite);
+    o = trigger->unkC0;
+    o->flags = (o->flags & ~4) | 8;
```
jiangzhengwenjz: Same as above

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+
+static void sub_080230DC(struct CutsceneTrigger5 *trigger)
+{
+    if (++trigger->unkB4 > 0x77)
+        trigger->obj2.unk78 = sub_08023100;
+}
+
+static void sub_08023100(struct CutsceneTrigger5 *trigger)
+{
+    gBldRegs.bldCnt = 0xBF;
```
jiangzhengwenjz: use `BLDCNT_*`

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+        o->unk3C = 0;
+        o->unk3E = 0;
+        o->sprite.unk8 = (o->sprite.unk8 & ~0x3000) | 0x1000;
+        n++;
+    }
+    PlaySfx(&trigger->obj2.base, SE_DARK_MIND_WARPSTAR_ENTER);
+    trigger->obj2.unk78 = sub_08021984;
+}
+
+static void sub_08021984(struct CutsceneTrigger2 *trigger)
```
jiangzhengwenjz, line 663: GPT gave a much cleaner match:
```diff
@@ -13,7 +13,7 @@
     struct CutsceneTrigger2 *trigger2 = trigger;
     u16 i;
     u16 j = 0;
-    u16 allDone;
+    bool32 allDone;
 
     for (i = 0; i < 4; i++)
     {
@@ -36,24 +36,12 @@
             {
                 diff = o->x - kirby->base.base.base.x;
                 spd = (diff >> 8) * gUnk_082DEA6C[j][0];
-                if (spd < gUnk_082DEA7C[j][0])
-                {
-                    u16 lo = gUnk_082DEA7C[j][0];
-
-                    spd = lo;
-                }
-                else
-                {
-                    s32 hi = gUnk_082DEA8C[j][0];
-                    s32 tmp = spd;
-                    if (tmp > hi)
-                        tmp = hi;
-                    spd = tmp;
-                }
-                o->unk3C = spd;
+                o->unk3C = spd < gUnk_082DEA7C[j][0]
+                    ? gUnk_082DEA7C[j][0]
+                    : min(gUnk_082DEA8C[j][0], spd);
                 if (o->unk3C < diff)
                 {
-                    o->unk3C = -spd;
+                    o->unk3C = -o->unk3C;
                     done = FALSE;
                 }
                 else
@@ -63,14 +51,9 @@
             {
                 diff = kirby->base.base.base.x - o->x;
                 spd = (diff >> 8) * gUnk_082DEA6C[j][0];
-                if (spd < gUnk_082DEA7C[j][0])
-                {
-                    u16 lo = gUnk_082DEA7C[j][0];
-                    spd = lo;
-                }
-                else if (spd > gUnk_082DEA8C[j][0])
-                    spd = gUnk_082DEA8C[j][0];
-                o->unk3C = spd;
+                o->unk3C = spd < gUnk_082DEA7C[j][0]
+                    ? gUnk_082DEA7C[j][0]
+                    : min(gUnk_082DEA8C[j][0], spd);
                 if (o->unk3C < diff)
                     done = FALSE;
                 else
@@ -85,14 +68,9 @@
             {
                 diff = o->y - kirby->base.base.base.y;
                 spd = (diff >> 8) * gUnk_082DEA6C[j][1];
-                if (spd < gUnk_082DEA7C[j][1])
-                {
-                    u16 lo = gUnk_082DEA7C[j][1];
-                    spd = lo;
-                }
-                else if (spd > gUnk_082DEA8C[j][1])
-                    spd = gUnk_082DEA8C[j][1];
-                o->unk3E = spd;
+                o->unk3E = spd < gUnk_082DEA7C[j][1]
+                    ? gUnk_082DEA7C[j][1]
+                    : min(gUnk_082DEA8C[j][1], spd);
                 if (o->unk3E < diff)
                     break;
                 o->unk3E = 0;
@@ -101,24 +79,13 @@
             {
                 diff = kirby->base.base.base.y - o->y;
                 spd = (diff >> 8) * gUnk_082DEA6C[j][1];
-                if (spd < gUnk_082DEA7C[j][1])
-                {
-                    u16 lo = gUnk_082DEA7C[j][1];
-                    spd = lo;
-                }
-                else
-                {
-                    s32 hi = gUnk_082DEA8C[j][1];
-                    s32 tmp = spd;
-                    if (tmp > hi)
-                        tmp = hi;
-                    spd = tmp;
-                }
-                o->unk3E = spd;
+                o->unk3E = spd < gUnk_082DEA7C[j][1]
+                    ? gUnk_082DEA7C[j][1]
+                    : min(gUnk_082DEA8C[j][1], spd);
                 if (o->unk3E < diff)
                 {
                     done = FALSE;
-                    o->unk3E = -spd;
+                    o->unk3E = -o->unk3E;
                 }
                 else
                     o->unk3E = 0;
@@ -176,7 +143,7 @@
         }
         j++;
     }
-    allDone = 1;
+    allDone = TRUE;
     for (i = 0; i < 4; i++)
     {
         if (trigger2->unk294[i] != NULL && trigger2->unk2A4[i] <= 4)

```

include/cutscene_trigger.h
```diff
@@ -0,0 +1,80 @@
...
+    /* 0x0B4 */ s16 unkB4;
+    /* 0x0B8 */ struct EffectObject *unkB8;
+    /* 0x0BC */ struct EffectObject *unkBC;
+    /* 0x0C0 */ struct EffectObject *unkC0;
+    /* 0x0C4 */ struct Unk_02022930_0 *unkC4;
+    /* 0x0C8 */ u8 unkC8;
+    /* 0x0C9 */ u8 unkC9;
+    /* 0x0CA */ s8 unkCA;
+    /* 0x0CB */ s8 unkCB;
+    /* 0x0CC */ u8 unkCC;
```
jiangzhengwenjz: Probably `bool8`? Please verify and make changes accordingly and use `TRUE`/`FALSE` if it's true.

jiangzhengwenjz (changes requested):

src/cutscene_trigger.c
```diff
@@ -0,0 +1,1953 @@
...
+            u16 angle;
+
+            if (++kirby->animationIndex > 0x59)
+                kirby->animationIndex = 0x4A;
+            spd = trigger->unkB4 / 2; // the u8 t staging and clamp arm order are required for matching
+            if (spd < 0xC8)
+                t = spd;
+            else
+                t = 0xC8;
+            spd8 = t;
```
jiangzhengwenjz: > the u8 t staging and clamp arm order are required for matching

This is wrong, according to GPT. `spd8 = min(trigger->unkB4 / 2, 0xC8);` matches

include/cutscene_trigger.h
```diff
@@ -0,0 +1,80 @@
...
+}; /* size = 0xEC */
+
+extern const u16 gUnk_082DE9FC[][2];
+extern const u16 gUnk_082DEA24[][2];
+extern const s16 gUnk_082DEA4C[4][2];
+extern const s16 gUnk_082DEA5C[4][2];
+extern const s16 gUnk_082DEA6C[][2];
+extern const s16 gUnk_082DEA7C[][2];
+extern const s16 gUnk_082DEA8C[][2];
+extern const s16 gUnk_082DEA9C[];
```
jiangzhengwenjz: GPT:
```
DEA6C = 0x10 -> s16 [4][2]
DEA7C = 0x10 -> s16 [4][2]
DEA8C = 0x10 -> s16 [4][2]
DEA9C = 0x08 -> s16 [4]
```
So the outer extent is definite and please fill the number in.

include/cutscene_trigger.h
```diff
@@ -0,0 +1,80 @@
...
+    /* 0x0CC */ u8 unkCC;
+    /* 0x0CE */ s16 unkCE[4][2];
+    /* 0x0DE */ u8 unkDE[4];
+    /* 0x0E2 */ u8 unkE2[4];
+    /* 0x0E6 */ s16 unkE6;
+    /* 0x0E8 */ s16 unkE8;
+}; /* size = 0xEC */
+
+extern const u16 gUnk_082DE9FC[][2];
+extern const u16 gUnk_082DEA24[][2];
```
jiangzhengwenjz: GPT suggested
```c
struct CutsceneAnimPair
{
    u16 animId;
    u16 variant;
};

extern const struct CutsceneAnimPair gUnk_082DE9FC[10];
extern const struct CutsceneAnimPair gUnk_082DEA24[10];
```
so we can have:
```c
animId = gUnk_082DE9FC[9].animId;
variant = gUnk_082DE9FC[9].variant;
```

Please change accordingly if it matches.

jiangzhengwenjz (changes requested):
GPT has a few more comments.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Exported the remaining seven distinct handlers and added declarations in their owning headers: BonusSetFunc, sub_08123780, WaddleDee37ChooseXSpeed, sub_080AB720, sub_080B6A54, sub_080C0CBC, and sub_080C2B28. They account for 18 entries in the merged baseline; the earlier count conflated table entries with distinct functions. The final ELF audit confirms that all 217 non-null entries now point to global Thumb functions at their original addresses. All six ROM checks pass.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Added the local TODO beside the switch. It identifies subtype 3 and values above 6 as leaving size/dtor uninitialized, preserves the original path, and explicitly leaves other runtime templates unresolved. The PR description now corrects the earlier overstatement about reachability.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Rescanned and formatted the whole file, including the over-indented bodies in sub_08020A90, its second switch, and both under-indented bodies in sub_08022770. Also removed a stray blank line during the independent follow-up audit. git diff --check and all six ROM comparisons pass.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Fixed the block body and surrounding condition formatting as part of the whole-file indentation pass. All six ROM comparisons pass.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Merged current master at 52fc58f in d2c3848, then adapted all cutscene Kirby accesses to the current layout and the existing palette macro signature. No history rewrite. The final clean build passes all six ROM comparisons.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Replaced the three literals with KIRBY_ABILITY_SLEEP, KIRBY_ABILITY_UFO, and KIRBY_ABILITY_CUPID. Verified their definitions and checked the other ability comparisons in this file. Instructions and relocations are unchanged.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Removed xyTable and use gUnk_082DEA4C directly. Tested this removal independently: the complete cutscene object's instructions and relocations are unchanged.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Removed xyTable2 and use gUnk_082DEA5C directly. Tested this removal independently: the complete cutscene object's instructions and relocations are unchanged.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Removed animTable, including its redundant loop assignment, and use gUnk_082DE9FC directly. Tested this removal independently before the record-type conversion; instructions and relocations are unchanged.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Both UpdateSaveBufferByOffset calls now use SAVE_BUFFER_TYPE_WORLD_PROPS, whose enum value is 1. Instructions and relocations are unchanged.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Updated this call as well to SAVE_BUFFER_TYPE_WORLD_PROPS; both branches retain the existing save-slot selection and match the ROM.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Current master already provides the variant parameter. After merging it, replaced this block and the other equivalent palette blocks with Macro_081050E8, preserving each condition and variant. Checked the replacements individually against the matching object; instructions and relocations are unchanged.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Replaced the second palette block with Macro_081050E8 using animId2 and variant2. Checked it independently; instructions and relocations are unchanged.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Removed flags and used the two |= statements. Tested this change independently; the complete cutscene object's instructions and relocations are unchanged.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Fixed the indentation of this block and checked the complete function and file for the same problem. Final clean ROM comparisons pass.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Fixed the indentation of this block along with the adjacent state-machine blocks. Final clean ROM comparisons pass.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Applied the two compound assignments here and in the second effect setup, then checked analogous combined flag assignments throughout this file. Each of the eight combined-update sites was tested independently and kept identical instructions and relocations.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Applied the same &= / |= form to this second effect setup and verified it independently. All six final ROM comparisons pass.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Replaced 0xBF with BLDCNT_TGT1_ALL | BLDCNT_EFFECT_LIGHTEN. These definitions preserve the exact register bits; instructions and relocations are unchanged.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Applied the cleaner form across all four clamps, including bool32 allDone, TRUE initialization, and negation of the stored speed fields. The earlier claim that the lo/hi/tmp locals were necessary was incorrect. This combined function rewrite matches exactly in the instruction/relocation comparison, and all six final ROM comparisons pass. An independent source/assembly review also checked the signed bounds and direction changes.

include/cutscene_trigger.h
jackmgaller: Fixed in the updates through 233e6e8. Verified subtype 5's accesses: zero initialization, TRUE when camera offsets are applied, FALSE when restored, and a Boolean test before pending restoration. Changed only that field to bool8 with TRUE/FALSE; subtype 0's unrelated unkCC bitmask remains u16. Layout and ROM bytes are unchanged.

src/cutscene_trigger.c
jackmgaller: Fixed in the updates through 233e6e8. Confirmed: spd8 = min(trigger->unkB4 / 2, 0xC8) matches. Removed spd, t, and the incorrect matching comment. Instructions and relocations are unchanged. The independent review also compared old/new byte-conversion behavior over all 65,536 signed-16 counter values and both direction cases.

include/cutscene_trigger.h
jackmgaller: Fixed in the updates through 233e6e8. Added all four explicit extents. data/data_14.s gives 0x10 bytes each for DEA6C/DEA7C/DEA8C and 8 bytes for DEA9C, consistent with the four-element loops and declared s16 widths. Verified the original symbol addresses and final ROM match.

include/cutscene_trigger.h
jackmgaller: Fixed in the updates through 233e6e8. Changed both tables to const CutsceneAnimPair[10], with u16 animId and variant at offsets 0 and 2, and updated every consumer to use the named fields. Each original table occupies 0x28 bytes. The type conversion was tested independently and preserves the complete object's instructions and relocations.

jiangzhengwenjz (approved):

jiangzhengwenjz:
```
1424856 total bytes of code
    1291564 bytes of code in src (90.6452%)
    122404 bytes of code in asm (8.5906%)
        115524 bytes of code remaining to be decompiled
```

jiangzhengwenjz:
```
41768 total symbols
    7015 symbols documented (16.7952%)
    34753 symbols undocumented (83.2048%)

14512864 total bytes of data

9868500 bytes of data in 29335 baserom incbins (67.9983%)
```

